PR #1, opened 2020-03-19
Feature/form improvements

mietcls:


tests/integration/components/au-label-test.js
```diff
@@ -0,0 +1,26 @@
...
+    // Set any properties with this.set('myProperty', 'value');
+    // Handle any actions with this.set('myAction', function(val) { ... });
+
+    await render(hbs`<AuLabel />`);
+
+    assert.equal(this.element.textContent.trim(), '');
+
+    // Template block usage:
+    await render(hbs`
+      <AuLabel>
```
Wolfr: <3

Wolfr:
Looks good to me.

Wolfr:
Did you manually create that integration test?

Is it a good practice, why did you do it? I tested how to generate Ember Octane components on Thursday; but when I generate, no tests are made.

mietcls:
We do it automatically, for instance with the command `edi ember g route route-name` – I do not understand why it did not do that with you? Is there a difference in stack @madnificent ?

Wolfr:
I use `ember generate component my-component -gc`.

madnificent:
Not sure what the state was before Octane landed officially, but `--gc` and `-gc` will currently generate a glimmer component and it should create a test for you too.  It's a stub test which simply tests whether the component crashes by default.

Without `--gc` a template-only component will be generated.  This is a component which has no logical backing.  Such components are easier to optimize and faster to render.  If you can express everything you need in the HTML, you might as well go with it.

In both cases a test will be generated.  If you intend to write tests you can run and inspect them in the browser by visiting [localhost:4200/tests](http://localhost:4200/tests)

PR #6, opened 2020-03-31
Bugfix/input field value

erikap:
- Fix value binding on AuInput and AuTextarea components
- Resolve linting errors


erikap:
PR should be created on development branch

PR #9, opened 2020-04-02
Feature/alerts

mietcls:
Starting to build alerts. Not in scope:
- styled buttons
- picking icons
- floating alerts

Just "ember serve" and look at the alerts in the docs. Thanks!

### DIFF
--- a/addon/components/au-alert.js
+++ b/addon/components/au-alert.js
@@ -20,4 +20,6 @@ export default class AuAlertComponent extends Component {
       return "au-c-alert--small";
     return "";
   }
+
+  // @TODO: add icon switching functionality
 }
--- a/addon/components/au-alert.js
+++ b/addon/components/au-alert.js
@@ -3,21 +3,21 @@ import Component from '@glimmer/component';
 export default class AuAlertComponent extends Component {
   get alertSkin() {
     if (this.args.alertskin == "info")
-      return "--info";
+      return "au-c-alert--info";
     if (this.args.alertskin == "success")
-      return "--success";
+      return "au-c-alert--success";
     if (this.args.alertskin == "warning")
-      return "--warning";
+      return "au-c-alert--warning";
     if (this.args.alertskin == "error")
-      return "--error";
+      return "au-c-alert--error";
     return "";
   }
 
   get alertSize() {
     if (this.args.alertsize == "tiny")
-      return "--tiny";
+      return "au-c-alert--tiny";
     if (this.args.alertsize == "small")
-      return "--small";
+      return "au-c-alert--small";
     return "";
   }
 }
--- a/addon/components/au-alert.js
+++ b/addon/components/au-alert.js
@@ -2,28 +2,22 @@ import Component from '@glimmer/component';
 
 export default class AuAlertComponent extends Component {
   get alertSkin() {
-    if (this.args.alertskin == "info") {
+    if (this.args.alertskin == "info")
       return "--info";
-    }
-    if (this.args.alertskin == "success") {
+    if (this.args.alertskin == "success")
       return "--success";
-    }
-    if (this.args.alertskin == "warning") {
+    if (this.args.alertskin == "warning")
       return "--warning";
-    }
-    if (this.args.alertskin == "error") {
+    if (this.args.alertskin == "error")
       return "--error";
-    }
     return "";
   }
 
   get alertSize() {
-    if (this.args.alertsize == "tiny") {
+    if (this.args.alertsize == "tiny")
       return "--tiny";
-    }
-    if (this.args.alertsize == "small") {
+    if (this.args.alertsize == "small")
       return "--small";
-    }
     return "";
   }
 }
--- a/addon/components/au-alert.js
+++ b/addon/components/au-alert.js
@@ -1,0 +1,29 @@
+import Component from '@glimmer/component';
+
+export default class AuAlertComponent extends Component {
+  get alertSkin() {
+    if (this.args.alertskin == "info") {
+      return "--info";
+    }
+    if (this.args.alertskin == "success") {
+      return "--success";
+    }
+    if (this.args.alertskin == "warning") {
+      return "--warning";
+    }
+    if (this.args.alertskin == "error") {
+      return "--error";
+    }
+    return "";
+  }
+
+  get alertSize() {
+    if (this.args.alertsize == "tiny") {
+      return "--tiny";
+    }
+    if (this.args.alertsize == "small") {
+      return "--small";
+    }
+    return "";
+  }
+}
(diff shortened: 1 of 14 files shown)


Wolfr:
Very nice! I love to see the colors in action. My comments below.

## Bugs

* `au-c-alert` seems to be rendered twice e.g. the first example has these classes: `au-c-alert au-c-alert--info au-c-alert`

## Semantics

* h1 is probably not a good default for `au-c-alert__title`. I would simply make the title another paragraph, otherwise there might be conflicts with the heading order of the rest of the page (which would be esp. annoying for screen reader users)

## Code style/markup

* I would avoid using utillity classes in components (`au-u-margin-bottom` for the spacing on the paragraph: I would either expect the hook to render rich text text here with an `.au-c-content` div, or expect just a single paragraph here with its own class hook.

The latter example would mean there is no support for more than 2 lines of texts in alerts. I think I saw some examples of more intricate alerts in your screenshots before.

* In the example with actions, the buttons are in a markup pattern that could be improved This is what I see:

```
<p class="au-c-alert__message">
      <p class="au-u-margin-bottom">Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.</p>
      <button class="au-c-button   ">
  Primary action
</button>
      <button class="au-c-button  au-c-button--ghost ">
  Secondary action
</button>
    </p>
```

* Imo, a specific pattern is needed for button spacing (e.g. an `.au-c-button-toolbar` component)
* The actions could be in their own wrapper for example ".au-c-alert__actions"

## Content

* I would use actual lorem ipsum, not the lorem ipsum text from lorem ipsum website, with content length that reflects what an alert typically has

E.g.

```
(line 1)
Sunt in culpa qui officia deserunt mollit

(line 2)
Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam.
```

## Enhancements

* Closable button does not have a hover, active and focus state (except for the general focus ring)


---

A lot of feedback but - love this! A really good step forward I think.

addon/components/au-alert.js
```diff
@@ -0,0 +1,23 @@
...
+  get alertSkin() {
+    if (this.args.alertskin == "info")
+      return "--info";
+    if (this.args.alertskin == "success")
+      return "--success";
+    if (this.args.alertskin == "warning")
+      return "--warning";
+    if (this.args.alertskin == "error")
+      return "--error";
+    return "";
```
rafael-mindreau, line 13: Can a switch be a better candidate for this control structure?

addon/components/au-button.js
```diff
@@ -3,14 +3,14 @@ import Component from '@glimmer/component';
...
+      return "au-c-button--block";
     else
       return "";
   }
 
   get skinClass() {
-    if (this.args.error == "ghost")
-      return "au-c-input--ghost";
+    if (this.args.skin == "ghost")
+      return "au-c-button--ghost";
```
rafael-mindreau, line 13: Does this project adhere to braceless control structures? That's very odd, this is probably the first time I've seen it in this format.

rafael-mindreau: Expanding on what I meants:
```javascript
if (this.args.skin == "ghost") {
    return "au-c-input--ghost";
} else {
  ...
}
```

What I used to do a long time ago is to in-line all braceless structures like so:
```javascript
if (this.args.skin == "ghost") return "au-c-input--ghost";
```

I'm just mentioning this out of curiosity, as it's out of the scope of this specific issue

mietcls: Thanks for the suggestion. I am not sure what the best approach is here. I've seen a bunch of different ones, but I am not the expert here. Love to see what the best practice is :-)!

addon/components/au-alert.js
mietcls: No clue honestly. We still need to have a decent conversation on what the best approach is here. This pattern will be reused in many places.

mietcls:
> Very nice! I love to see the colors in action. My comments below.
> 
> ## Bugs
> * `au-c-alert` seems to be rendered twice e.g. the first example has these classes: `au-c-alert au-c-alert--info au-c-alert`
> 
👌 will fix

> ## Semantics
> * h1 is probably not a good default for `au-c-alert__title`. I would simply make the title another paragraph, otherwise there might be conflicts with the heading order of the rest of the page (which would be esp. annoying for screen reader users)
> 
👌 will fix

> ## Code style/markup
> * I would avoid using utillity classes in components (`au-u-margin-bottom` for the spacing on the paragraph: I would either expect the hook to render rich text text here with an `.au-c-content` div, or expect just a single paragraph here with its own class hook.
> 
Fixed the fact that there is a paragraph in a paragraph –and you can use multiple paragraphs.
Will not fix the au-u-margin-bottom thingy for now as it probably needs to be fixed on a more general level.

> The latter example would mean there is no support for more than 2 lines of texts in alerts. I think I saw some examples of more intricate alerts in your screenshots before.
> 
> * In the example with actions, the buttons are in a markup pattern that could be improved This is what I see:
> 
> ```
> <p class="au-c-alert__message">
>       <p class="au-u-margin-bottom">Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book.</p>
>       <button class="au-c-button   ">
>   Primary action
> </button>
>       <button class="au-c-button  au-c-button--ghost ">
>   Secondary action
> </button>
>     </p>
> ```
> 
> * Imo, a specific pattern is needed for button spacing (e.g. an `.au-c-button-toolbar` component)
> * The actions could be in their own wrapper for example ".au-c-alert__actions"
> 
Needs to be fixed later, button-group etc too.

> ## Content
> * I would use actual lorem ipsum, not the lorem ipsum text from lorem ipsum website, with content length that reflects what an alert typically has
> 
> E.g.
> 
> ```
> (line 1)
> Sunt in culpa qui officia deserunt mollit
> 
> (line 2)
> Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam.
> ```
> 
I hate normal lorem ipsum but you are right, there is no copyright on normal lorem ;-)

> ## Enhancements
> * Closable button does not have a hover, active and focus state (except for the general focus ring)
> 
> A lot of feedback but - love this! A really good step forward I think.
We need a larger discussion about buttons. There are a lot of button styles right now, the system could be more clear. Not in scope for me now.

PR #10, opened 2020-04-02
Feature/form improvements

mietcls:
- Starting to add loading states (ugly). Loading states need a better way to write the css for this I guess.
- Experimenting with error states on labels, and aligning style
- Add some documentation

I mostly need this to get started on form validation. Might not solve everything at once now.


mietcls:
Will merge this for demo purposes. We will later on fix:
- color usage (when to use neutral, when to use gray?)
- refactor the button logic

PR #20, opened 2020-04-13
Upgrade to appuniversum 0.0.7

Dietr:
- change color variables
- minor code style fixes


Dietr:
The editorconfig (ember defaults) states that .hbs files have no newline on the end.

PR #21, opened 2020-04-16
Feature/badge icon

mietcls:
Basic icon functionality. No hover states for action badge or sizes defined.


addon/components/au-badge.js
```diff
@@ -0,0 +1,17 @@
+import Component from '@glimmer/component';
+
+export default class AuBadgeComponent extends Component {
+  get skin() {
+    if (this.args.skin == "info")
```
Wolfr, line 5: I like to add code that reacts to not having a skin prop, like so `if (this.args.skin == "info" || !args.skin)` so there is a default badge that works, when no skin is provided.

I would consider this grey one to be the default:

image

What do you think? This way the implementation can be 

`<AuBadge @icon="info-circle" />` (can you self-close components?)

tests/dummy/app/templates/docs/atoms/au-badge.md
```diff
@@ -0,0 +1,23 @@
+# Badge
+
+---
+
+{{#docs-demo as |demo|}}
+  {{#demo.example name='au-Badge.hbs'}}
+    <AuBadge @icon="info-circle"></AuBadge>
+    <hr>
+    <AuBadge @icon="info-circle" @skin="info"></AuBadge>
```
Wolfr, line 9: Is it possible to self-close?

`<AuBadge @icon="info-circle" @skin="info" />`

addon/components/au-badge.js
mietcls: I have this one as the default (when no skin is added)
Screenshot 2020-05-06 at 15 15 40
:
Screenshot 2020-05-06 at 15 14 09

Which is solved in the css code
It's a different approach, but maybe not standard.
Screenshot 2020-05-06 at 15 14 50


What is hte

tests/dummy/app/templates/docs/atoms/au-badge.md
mietcls: Yes

addon/components/au-badge.js
Wolfr: I don't think it's necessary to do a `:not` in your CSS, it's better to not render the class in the first place.

This is all new terroritory for me so, take what I will write here with a grain of salt, I still have to write more components in this Octane style to find the ideal workflow.

I find with this kind of code, you should define every variant in your CSS as a modifier:

```
.c-my-component {}
.c-my-component--default {}
.c-my-component--variant-1 {}
.c-my-component--variant-2 {}
```

This way you don't have to override any styles in your modifier class.

And then when the component is called, and you do not provide an `@skin` prop and thus write something like:

```<MyComponent />```

It should render out the default:

```<div class="c-my-component c-my-component--default"></div>```

To do so in JS one would probably write something like

```

export default class MyComponent extends Component {
  get skin() {
    if (this.args.skin == "variant1")
      return "c-my-component--variant1";
    if (this.args.skin == "variant2")
      return "c-my-component--variant2";
    else
      return "c-my-component--default";
  }
}
```

PR #40, opened 2020-09-15
Remove overflow auto

aatauil:
This is the cause for the ember-power-select dropdown indent on the mandatendatabank homepage. Weirdly enough removing 'display: flex;' from 'au-c-main-header__title-group' also fixes this issue but causes the header to get messed up. Removing overflow:auto does not have any impact visuals and fixes the issue on all viewport sizes.

### DIFF
--- a/app/styles/ember-appuniversum/_c-main-container.scss
+++ b/app/styles/ember-appuniversum/_c-main-container.scss
@@ -42,5 +42,4 @@ $au-main-container-height: calc(100vh - 4.2rem) !default;
 
 .au-c-main-container__content--scroll {
   height: 100%;
-  overflow: auto;
 }


aatauil:
Maybe not remove remove overflow auto but I dont know what that css class does so I cant really test if it has to be overflow hidden or not

Dietr:
Removing the overflow auto breaks the way the main container component works. Can I test this issue somewhere? My first guess is that the 'au-c-main-header__title-group' is responsible so I would like to look into reworking this component.

aatauil:
I already though something like that would be the case.. I dont exactly know the best way for you to test it, I both cloned frontend-mandatendatabank & ember-appuniversum & performed an npm link to actually debug with live reload, this should also work for you I'm pretty sure

Dietr:
I did some tests on this. It seems like an issue with the ember-power-select and the scroll containers we are using in ember-appuniversum. The offset of the dropdown is equal to the scrollbar on the au-c-main-container--scroll element (which has overflow:auto to display scrollbars when needed). As soon as you start typing in the power-select search field the position is fixed. I have no solution yet but will create a new issue for this.

aatauil:
I see, have you taken into consideration that when removing ` display:flex` from `au-c-main-header__title-group` also resolves the issue ? The element having that class exists much higher in the dom tree so I don't immediately see how that could have an impact on the `c-main-container--scroll ` that is nested quite some levels deeper in the tree. I could be completely off though but I just want to give you a heads up on that as it pretty weird behavior from my experience.

Dietr:
See my comment in https://github.com/appuniversum/ember-appuniversum/issues/41

PR #52, opened 2020-10-08
Add date-picker component (Duetds)

aatauil:
DONE:
- Transformed Duetds date-picker into an ember-component 
- Filled out documentation with demo's and arguments
- Added error handling
- Modified the default values slightly

NOTES:
- Tested it 
- Should not interfere with existing code

TO-DO:
- Styling is default. Changes will probably need to be made, an example is when passing in the attribute disabled=true, there is no visual feedback to the date picker being disabled.
- Maybe change default formatting to something more suitable

ref: https://github.com/duetds/date-picker

### DIFF
--- a/addon/components/au-date-picker.hbs
+++ b/addon/components/au-date-picker.hbs
@@ -3,5 +3,5 @@
     <label for="date">{{@label}}</label>
   {{/if}}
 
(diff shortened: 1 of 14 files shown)


Dietr:
Nice work, I'll merge and create a custom theme for this

PR #54, opened 2020-10-14
Enable usage of the model attribute

abeforgit:
Disclaimer: change made with github editor, not tested (I tried, but couldn't figure out how do quickly test it, and there is really no reason this wouldn't work)


abeforgit:
Just tested it locally in the GN context, it works (and doesnt work without the modification)

PR #58, opened 2020-11-12
Dev/time picker (will need a review first see description)

aatauil:
Working time picker with keyboard support.
Made some radical changes, biggest being that you can not input numbers yourself. You can do this using keyboard up and down or the plus/minus buttons on the side. Reason being that supporting manual number input is very prone to errors and does not play well with showing double digits (e.g. 04 instead of just 4)

Dont really know if this is going to be problematic for user experience, as a compensation I can add support for incrementing/decrementing time value while one of the buttons is being held down.

Let me know what you think :)

btw the PR request in pointed towards development branch instead of master

### DIFF
--- a/addon/components/au-time-picker.hbs
+++ b/addon/components/au-time-picker.hbs
@@ -4,7 +4,7 @@
     <div class="au-time-picker_box">
       <label class="au-time-picker_label" for="">Hours</label>
       <div class="au-time-picker_input-wrapper">
-        <input class="au-time-picker_input" value={{this.hourValue}}  {{on 'keydown' (fn this.setTimeValue "hourValue")}} name="input-hour" id="input-hour"/>
+        <input class="au-time-picker_input" value={{format-time this.hourValue}}  {{on 'keydown' (fn this.setTimeValue "hourValue")}} name="input-hour" id="input-hour"/>
         <div class="au-time-picker_button-wrapper">
           <button type="button" aria-label="increment hours" {{on 'click' (fn this.increment "hourValue")}} aria-controls="input-hour"  class="au-time-picker_button">+</button>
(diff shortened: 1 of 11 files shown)


mietcls:
First thoughts on the manual input (I will have to really see how it works still, I do not have to check it out in depth right now):

Not being able to manually input the number will probably frustrate people that need to fill out forms on a daily base, as quickly as possible.
- Long press takes more time than just typing the number (if they discover it)
- If they don't discover long press, it would mean that if they want to add, for example, the number 30 they'll have to click 30 times

A dropdown could be another solution, but this also takes a lot more time than just typing the number.

In our previous time pickers I think @madnificent had a fix for the "04" problem. Do you have some pointers?

madnificent:
I don't have any pointers to it by hand.  But it should be feasible to fix.  The interesting problem arrives when the input can come from multiple sources (eg: button and text) and is further complicated when one of the sources is also the output (eg: the number you type should be consumed, but should be updated too).

It is an interesting problem to work yourself through as it requires you to have some "base data" and know how to update that base data based on certain events.  Assuming that is the current state, and that it's a soft or hard requirement to have digits input (easy to imagine indeed), I'd love to pair on it with @aatauil.  I hope we can find the time for it.

aatauil:
@madnificent Those are the exact issues I am having with manual input ( and some small other stuff e.g. focusout events after input ). Would appreciate a ping so we can pair whenever you have some spare time

aatauil:
@miet @Dietr I updated my previous message but I think you probably did not receive any notifications of it. I made it work with user input and formatting and added a 'now' button that updates to current time. Can you review this and see if this has all we need? Thankss

Dietr:
@aatauil Looks good, seems to work as expected in my tests. I will add the styling to this and include it in the next release.

PR #60, opened 2020-11-23
Feature/update modal functionality

abeforgit:
Allows the modal to be controlled by the calling component. Also sets inert state on either the modal or the rest of the document for better a11y.

This pr includes an ember version bump to the latest lts release. The reason for this is the use of the `{{in-element}}` helper. Codemods have been run, and tests have also been updated.


abeforgit:
Note there is one test still breaking, I will make a separate pr for this as it involves a change in the component code

PR #62, opened 2020-11-23
Add time-picker & date-picker functionality

aatauil:
**Time-picker**
- You can set values through attributes
- onChange attribute has been added that returns an object with all time values when a change is made
- tested seperatly
- updated docs with changes

**Date-picker**
- onChange attribute has been added that returns the element object including date value when a change is made
- tested seperatly
- updated docs with changes

### DIFF
--- a/addon/components/au-date-picker.hbs
+++ b/addon/components/au-date-picker.hbs
@@ -3,5 +3,5 @@
     <AuLabel for="date">{{@label}}</AuLabel>
   {{/if}}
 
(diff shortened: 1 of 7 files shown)


cecemel (commented):
Hi, thanks for the upates.
Some extra remarks:

- the interface of onChange should be cleaner. Now it returns an event from duet, but actually, I would like to have abstraction.
  What if you change duet as library. All code using the component will need an update too.
  To keep it simple the provided callback would be this signature:
  ```
  @action
  yourClientsCallbackOnChange(dateString){
  //date string formatted as
  //Must be in IS0-8601 format: YYYY-MM-DD. -> this to keep the setting and getting uniform with what duet expects
  //Tip you can cast POJO dataobject like new Date('12/12/2012').toISOString().split('T')[0];
  //ISO-8601 is a standard, so independen of the used library
  }
  ```

PR #84, opened 2021-03-15
Improvement radiobuttons

aatauil:
- Convert to ember input component
- replace @value by @checked
- pass in selected state, value & event context to callback

### DIFF
--- a/addon/components/au-control-radio.hbs
+++ b/addon/components/au-control-radio.hbs
@@ -1,5 +1,5 @@
 <label for="{{@identifier}}" class="au-c-control au-c-control--radio {{this.disabled}}">
-  <input name="{{@name}}" id="{{@identifier}}" value="{{@value}}" disabled={{@disabled}} type="radio" class="au-c-control__input" {{on "change" this.onChange}} ...attributes>
+  <Input class="au-c-control__input" id="{{@identifier}}" name="{{@name}}" disabled={{@disabled}} @type="radio" @checked={{@checked}} {{on "change" this.onChange}} ...attributes />
   <span class="au-c-control__indicator"></span>
   {{@label}}
 </label>
--- a/addon/components/au-control-radio.js
+++ b/addon/components/au-control-radio.js
@@ -12,7 +12,7 @@ export default class AuControlRadio extends Component {
   @action
   onChange(event) {
     if (this.args.onChange) {
-      this.args.onChange(event.target.checked, event);
+      this.args.onChange(event.target.checked, event.target.value, event);
     }
   }
 }
--- a/addon/components/au-control-radio.js
+++ b/addon/components/au-control-radio.js
@@ -12,7 +12,7 @@ export default class AuControlRadio extends Component {
   @action
   onChange(event) {
     if (this.args.onChange) {
-      this.args.onChange(event.target.value);
+      this.args.onChange(event.target.checked, event);
     }
   }
 }
--- a/tests/dummy/app/templates/docs/atoms/au-form-radio.md
+++ b/tests/dummy/app/templates/docs/atoms/au-form-radio.md
@@ -47,4 +47,4 @@
 | `@name` | Groups radio buttons, to make sure their selected statuses are related  | `String` | - |
 | `@identifier` | Makes sure the label is linked to the radio button and clicking it will only activate the radio button you clicked.  | `String` | - |
 | `@checked` | Used to set/get checked state of component  | `Boolean` | - |
-| `@onChange`| Expects a function that gets called when the radio button state changes. The function receives the radio button state & event context as parameters. | `Function` | | - |
+| `@onChange`| Expects a function that gets called when the radio button state changes. The function receives the radio button **state**, **value** & **event context** as parameters. | `Function` | | - |
--- a/tests/dummy/app/templates/docs/atoms/au-form-radio.md
+++ b/tests/dummy/app/templates/docs/atoms/au-form-radio.md
@@ -9,10 +9,10 @@
     <div class="au-c-form">
       <ul>
         <li>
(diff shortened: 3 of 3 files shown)


addon/components/au-control-radio.hbs
```diff
@@ -1,5 +1,5 @@
 <label for="{{@identifier}}" class="au-c-control au-c-control--radio {{this.disabled}}">
-  <input name="{{@name}}" id="{{@identifier}}" value="{{@value}}" disabled={{@disabled}} type="radio" class="au-c-control__input" {{on "change" this.onChange}} ...attributes>
+  <Input class="au-c-control__input" id="{{@identifier}}" name="{{@name}}" disabled={{@disabled}} @type="radio" @checked={{@checked}} {{on "change" this.onChange}} ...attributes />
```
nvdk, line 2: you seem to have lost @value here

addon/components/au-control-radio.js
```diff
@@ -12,7 +12,7 @@ export default class AuControlRadio extends Component {
   @action
   onChange(event) {
     if (this.args.onChange) {
-      this.args.onChange(event.target.value);
+      this.args.onChange(event.target.checked, event.target.value, event);
```
nvdk, line 15: by changing the order of the parameters here, you've introduced a breaking change.

tests/dummy/app/templates/docs/atoms/au-form-radio.md
```diff
@@ -9,10 +9,10 @@
     <div class="au-c-form">
       <ul>
         <li>
-          <AuControlRadio @label="Ja" @name="ja-nee" @value="ja" @identifier="ja" />
+          <AuControlRadio @label="Ja" @name="ja-nee"  @identifier="ja" value="ja" />
```
nvdk, line 12: moving from `@value` to `value` is a breaking change

addon/components/au-control-radio.hbs
aatauil: @value is not really needed as it is static in the radio-button case, you can use the regular 'value' (also using @checked & @value in the same input gives you a linting error. From what I learned is that the Ember Input component gives you the power of two-way binding for the '@checked' for example. 

https://github.com/emberjs/ember.js/blob/v3.25.1/packages/%40ember/-internals/glimmer/lib/components/input.ts#L77-L79

addon/components/au-control-radio.js
aatauil: Yes, that did come to mind when I coded this. I asked Dieter if the radio button was used somewhere critical and other then the contact form nothing was else was mentioned, I can change the order around so value gets passed in first :)

tests/dummy/app/templates/docs/atoms/au-form-radio.md
aatauil: Is it possible to remove the @ on every instance? it will still work the same when this is done, I prioritized the "@checked" over "@value"  since that one is more likely to need two-way binding rather the "value" (linting error doesnt like you to use both @checked & @value in an Input component)

addon/components/au-control-radio.hbs
nvdk: I tend to view the two-way binding of `@checked` as a legacy feature that will be deprecated sooner or later. the value attribute of a radio button is used to identify which radio button is checked. 

This allows you to use the same callback for all radio buttons for example

aatauil: Yeah I think that might be a better approach, I actually thought that that was the part in radio buttons that needed fixing ? I think it got communicated wrongly then. Thought there was need for a way to get both the checked state & its value after an action

nvdk:
I think we use a radio button very differently. IMO whether a radio button is checked or not doesn't matter at all. You want to know which one was selected and you use the value attribute to determine which one that is. I feel like these changes push using a radio button very much like a checkbox , which feels off to me.


On 16 March 2021 17:31:20 CET, Achraf Atauil ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -1,5 +1,5 @@
> <label ***@***.***}}" class="au-c-control au-c-control--radio {{this.disabled}}">
>-  <input ***@***.***}}" ***@***.***}}" ***@***.***}}" ***@***.***}} type="radio" class="au-c-control__input" {{on "change" this.onChange}} ...attributes>
>+  <Input class="au-c-control__input" ***@***.***}}" ***@***.***}}" ***@***.***}} @type="radio" @***@***.***}} {{on "change" this.onChange}} ...attributes />
>
***@***.*** is not really needed as it is static in the radio-button case, you can use the regular 'value' (also using @checked & @value in the same input gives you a linting error. From what I learned is that the Ember Input component gives you the power of two-way binding for the ***@***.***' for example. 
>
>https://github.com/emberjs/ember.js/blob/v3.25.1/packages/%40ember/-internals/glimmer/lib/components/input.ts#L77-L79
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/appuniversum/ember-appuniversum/pull/84#discussion_r595340783
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

addon/components/au-control-radio.hbs
nvdk: given the discussion in our #ember channel and on https://discuss.emberjs.com/t/ember-octane-and-2-way-binding-what-are-you-recommend/18092/15 I'd avoid wrapping the `Input` component and rather use the native `input` element.

nvdk: but perhaps we should discuss the intend of this PR first :)

aatauil:
Closing: no need to modify radio buttons

PR #93, opened 2021-04-26
`<AuDatePicker>` improvements

### LINKED ISSUE #90: [AuDatePicker] Suggestions / issues
I'm integrating the AuDatePicker component in GN and I noticed some things that can be improved:

#### Integration issues
At the moment the `<duet-date-picker>` component doesn't actually work in other projects.  The html is rendered, but the web component logic isn't being registered / executed. This addon uses `ember-cli-stencil` to make stencil web components work automagically, but it is installed as a devDependency which means it's not installed in the host app. It might be possible to simply move it to the dependencies but it would have to be tested.

An alternative could be to create an initializer which does the initialization. The code would look similar to [this commit](https://github.com/lblod/frontend-gelinkt-notuleren/pull/89/commits/d77baa6df3be6017e4cbab839b11a2778e29663e). I think this approach is better since the ember-cli-stencil addon does some very dynamic things which I suspect won't be compatible with Embroider. Since the manual version is so simple I think it makes sense to just sidestep that problem altogether.

#### Suggested API changes:
- The AuDatePicker expects an ISO 8601 formatted date as a value because it is passed through as-is to the `<duet-date-picker>` component. Since Ember supports arguments of different types than string, I think it would make more sense to support `Date` instances as well (or only). That way the consuming app doesn't have to convert dates to strings manually.
- The same goes for the value that gets passed into the callback function. Returning a `Date` instance to update a date or undefined / null to clear it would be a nicer DX. 

#### Issues:
- Typing in the input field doesn't update the calendar when opened. This does work on the `<duet-date-picker>` demo page although I haven't looked into the details yet.
- Clearing the input field throws an exception because the [`callBackParent` action](https://github.com/appuniversum/ember-appuniversum/blob/c040b0f6292c255fb85a14d1c5d30a726510474a/addon/components/au-date-picker.js#L87) doesn't handle that yet.
- Translations are set when the component is created so to change the translations the component needs to be destroyed first. I think it would be better to convert this logic into a getter so it gets updated if the `localization` argument changes. Not super important since it will probably not be a problem in practice, but it would make the component more solid.

PR for reference: https://github.com/lblod/frontend-gelinkt-notuleren/pull/89

Windvis:
This implements the changes suggested in #90.

- replaces ember-cli-stencil with a manual component registration which should fix the problem that the date picker doesn't work when embedded in an actual app.
- adds support for Date instances to the relevant arguments and onChange callback
- changes the default adapter parser to match the placeholder. This "fixes" the issue where typing the date manually doesn't update the date picker when opened. The problem was that the placeholder didn't match the parser so the typed date was never registered as a change.
- fixes the exception when clearing the input field
- code refactoring :nerd_face: 

Todo:
- ~~commit cleanup~~
- ~~documentation updates~~
- ~~tests~~ Testing web components is really painful. Timing issues everywhere..

Closes #90


addon/components/au-date-picker.js
```diff
@@ -1,93 +1,142 @@
...
+  closeLabel: 'Close window',
+  keyboardInstruction: 'You can use arrow keys to navigate dates',
+  calendarHeading: 'Choose a date',
+  dayNames: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
+  monthNames: ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'],
+  monthNamesShort: ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'],
+};
 
 export default class AuDatePickerComponent extends Component {
+  @isoDate value;
```
Windvis, line 35: `@isoDate` might not be the clearest name but couldn't think of something else.

addon/components/au-date-picker.hbs
```diff
@@ -1,7 +1,20 @@
-<div class="au-date-picker">
+<div class="au-date-picker" data-test-au-date-picker>
   {{#if @label}}
-    <AuLabel for="date">{{@label}}</AuLabel>
+    <AuLabel for={{this.id}} data-test-au-date-picker-label>{{@label}}</AuLabel>
```
nvdk, line 3: I seem to remember there being issues with html attributes without values in glimmer, I think only specific attributes can exist without a value (such as checked, disabled, etc). Unfortunately I can't find the issue immediately.

addon/components/au-date-picker.js
nvdk: perhaps `dateRepresentation` or `isoDateRepresentation`?

addon/components/au-date-picker.hbs
Windvis: What kind of issues? I can't recall anything like that :smile:.

addon/components/au-date-picker.js
```diff
@@ -1,93 +1,142 @@
...
-    this._assignAdapter(args);
+    return {
+      ...DEFAULT_ADAPTER,
+      ...this.args.adapter,
+    };
   }
 
-  @action _assignLocalization(args){
+  get id() {
+    return this.args.id ? this.args.id : guidFor(this);
```
nvdk, line 53: I guess this would also benefit from the unique-id polyfill that you've introduced in GN?

tests/integration/components/au-date-picker-test.js
```diff
@@ -1,26 +1,121 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render } from '@ember/test-helpers';
+import { click, render, settled, pauseTest, waitFor, waitUntil } from '@ember/test-helpers';
```
nvdk, line 3: some unused imports here I think

addon/components/au-date-picker.js
Windvis: Yea, but then this addon would need to ship that polyfill as well (and do the conditional logic in the template), so I decided against it since it's simple enough. It might be worth it if other components do something similar, but I would have to check.

addon/components/au-date-picker.hbs
nvdk: since I can't find it again, I guess we can assume this is resolved :+1:

nvdk (approved):
Looks good, did not see any oddities in the code and seems to be working well. It's unfortunate that we need the `await webComponentRender` in tests as that will make the test suite considerably slower, but I assume you've investigated alternatives and this is the best way forward for now.

Might benefit from another review, as I was the one pushing for supporting both string isoDate and date object. In my opinion the amount of added complexity seems limited.

tests/integration/components/au-date-picker-test.js
```diff
@@ -1,26 +1,121 @@
...
+
+  test('it uses the provided id to link the label to the input', async function(assert) {
+    await render(hbs`
+      <AuDatePicker @label="Choose a date" @id="label-test" />
+    `);
+    await webComponentRender();
+
+    let labelFor = this.element.querySelector('[data-test-au-date-picker-label]').getAttribute('for');
+    assert.equal(labelFor, 'label-test');
+    // await pauseTest()
```
Windvis, line 26: I forgot to remove this one.

tests/integration/components/au-date-picker-test.js
```diff
@@ -1,26 +1,121 @@
...
+    await datePickerOpened();
+
+    // Another hack to make sure the date picker buttons have actual values..
+    await waitUntil(() => {
+      return this.element.querySelector('.duet-date__day').children[0].innerText !== '';
+    }, { timeout: 2000 })
+
+    let availableDateButtons = this.element.querySelectorAll('.duet-date__day');
+    let targetDateButton = [...availableDateButtons].find((date) => {
+      // debugger;
```
Windvis, line 99: Same here.

Windvis:
Yes, testing web components isn't really straight forward since Ember doesn't know about them (which is way I kept it basic). I asked in the discord if there are any better ways of handling things, but I haven't received a response yet. I'll tweak things once I learn of better patterns.

I don't think it will make the test suite that much slower though? I think the 2000ms is the maximum amount of time it will wait before throwing an error but it doesn't seem to take nearly that long, I just added it like that since it was the default in the helper docs, and I wanted to be safe as well :smile:. I can always lower it if you want though. Or is there another reason why you think it will slow down the tests a lot?

nvdk:
> Yes, testing web components isn't really straight forward since Ember doesn't know about them (which is way I kept it basic). I asked in the discord if there are any better ways of handling things, but I haven't received a response yet. I'll tweak things once I learn of better patterns.
> 
> I don't think it will make the test suite that much slower though? I think the 2000ms is the maximum amount of time it will wait before throwing an error but it doesn't seem to take nearly that long, I just added it like that since it was the default in the helper docs, and I wanted to be safe as well smile. I can always lower it if you want though. Or is there another reason why you think it will slow down the tests a lot?

ah no, I thought it would be close to 2sec given that you provided that as a maximum. I guess it's better to provide a larger timerange given we'd rather not have false negatives.

Windvis:
Someone in the discord linked me to https://github.com/emberjs/ember-test-waiters which could clean the tests up a bit. The `webComponentRender` would still be needed though since `<duet-date-picker>` doesn't seem to have a callback / event for initialization / updates..

addon/components/au-date-picker.js
Windvis: Renamed to `isoDateRepresentation`.

Windvis:
@Dietr Should be good to go now!

PR #96, opened 2021-04-30
Deprecation warning fixed in v2.0.1 release

aatauil:
No need to use my fork anymore


Dietr:
@aatauil Looks like v2.0.1 hasn't been released yet (https://www.npmjs.com/package/ember-data-table)

aatauil:
Had not noticed that. I asked for it to be published, I will let you know when this happened :)

PR #97, opened 2021-05-05
Feature/card standout

mietcls:
Could anyone review if the pattern (edit-save content of a card) I created at the bottom for cards should be there, or in a different place?


Dietr:
I would expect the edit/save action inside the card for when there is a design without a title of the card. Or I would document this in under the pattern section. Now it's a very specific case that's documented inside the card component where the @standout is the only feature being added.

mietcls:
Discussion:
- review how we can hide the header when not needed
- review what we'll do with the fact that this is an article and probably needs a title
- review the pattern of having a title on top

PR #104, opened 2021-05-12
Update link to for embroider

bgantzler:


Windvis:
Hey @bgantzler thanks for looking into this!

I think I'm just going to change the implementation so that the component helper isn't needed. I also think this should simply "just work" with Embroider without having to pass the class. I consider this a "bug" in Embroider until it's confirmed by the Embroider team that this is the intended behavior. I just have to find some time to create an issue and reproduction :smile:.

Windvis:
I removed the need for the component helper in https://github.com/appuniversum/ember-appuniversum/pull/106, so I think this PR can be closed.

Thanks again for looking into this though!

PR #109, opened 2021-05-21
Rename arguments + icon logic

aatauil:
- make argument more inline with rest of app
- add defaults for open/closed icons => `open = down` :: `closed == up`
- add/test option to set icons to false
- update docs

Note: setting icon to false doesnt remove it from html but just doesnt assign an icon so it displays nothing 

@Windvis would also like you to take a quick look at it :)

### DIFF
--- a/addon/components/au-accordion.hbs
+++ b/addon/components/au-accordion.hbs
@@ -4,24 +4,24 @@
       <div>
         <AuButton @skin="tertiary" aria-hidden="true" focusable="false" aria-expanded="{{if this.openAccordion "true" "false"}}">
           {{#if @hideText}}
-            <span class="au-u-hidden-visually">{{@accordionButtonLabel}}</span>
+            <span class="au-u-hidden-visually">{{@label}}</span>
           {{else}}
-            {{@accordionButtonLabel}}
+            {{@label}}
           {{/if}}
         </AuButton>
         <p>
-          {{@accordionSubTitle}}
+          {{@subTitle}}
         </p>
       </div>
     </AuToolbarGroup>
     <AuToolbarGroup>
       {{#if this.accordionOpen}}
-        <AuIcon @icon="{{@accordionIconOpen}}" @alignment="left" @size="large" @ariaHidden="true" />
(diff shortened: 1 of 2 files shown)


addon/components/au-accordion.hbs
```diff
@@ -4,24 +4,24 @@
...
-            <span class="au-u-hidden-visually">{{@accordionButtonLabel}}</span>
+            <span class="au-u-hidden-visually">{{@label}}</span>
           {{else}}
-            {{@accordionButtonLabel}}
+            {{@label}}
           {{/if}}
         </AuButton>
         <p>
-          {{@accordionSubTitle}}
+          {{@subTitle}}
```
Windvis, line 13: I would go with `subtitle` since it's one word in English :smile:.

addon/components/au-accordion.hbs
```diff
@@ -4,24 +4,24 @@
...
         </p>
       </div>
     </AuToolbarGroup>
     <AuToolbarGroup>
       {{#if this.accordionOpen}}
-        <AuIcon @icon="{{@accordionIconOpen}}" @alignment="left" @size="large" @ariaHidden="true" />
+        <AuIcon @icon="{{if @iconOpen @iconOpen 'nav-up'}}" @alignment="left" @size="large" @ariaHidden="true" />
```
Windvis, line 19: I would do this in a getter in the backing class. That way you can also take the open state into account and simplify this whole if / else structure.

> Note: setting icon to false doesnt remove it from html but just doesnt assign an icon so it displays nothing

I think passing in `{{false}}` would return `nav-up` instead of an empty icon? I'm guessing you meant the string "false" and that would be true for any invalid icon name I suppose. 

Good point though, should it be possible to not show the icons? In that case an extra "showIcons" flag (which defaults to true) might be better? (and not show the icon component at all)

Windvis (commented):
Nice! Some minor questions / suggestions, but I like the improvements so far!

addon/components/au-accordion.hbs
aatauil: ah I was really thinking about that! but saw that it was already camel case so was like yeah it should probably be like that I guess? I agree, will change it right away

Dietr:
Included this in the storybook PR

Issue #88, opened 2021-04-09
Support all LinkTo arguments on the LinkTo wrapper components


Windvis:
`<AuNavigationLink>` and `<AuLink>` don't support all routing related arguments that LinkTo supports, which means that you sometimes have to use a LinkTo and add the css classes yourselves if you want to target a specific route.

Missing arguments for `<AuLink>`
- ~~`@models`~~
- `@query`

Missing arguments for `<AuNavigationLink>`
- ~~`@model`~~
- ~~`@models`~~
- `@query`

Windvis:
I also think it would be better to rename the `@linkRoute` argument to `@route` since that's what `<LinkTo>` uses and most people will try to use if they don't check the docs. I think aliasing and deprecating `@linkRoute` is the way to go, so apps aren't forced to update everything when they want to update ember-appuniversum.

abeforgit:
I see this has been merged into dev, any chance we can get this into GN asap?

PR #115, opened 2021-06-09
Deprecate the @linkRoute argument for the link components

Windvis:
[As mentioned before](https://github.com/appuniversum/ember-appuniversum/issues/88#issuecomment-844026558) I think `@route` is a better and more intuitive name for that argument since that's what `<LinkTo>` uses.

Not a breaking change since `@linkRoute` still works (but it will show a deprecation).

### DIFF
--- a/addon/components/au-link.js
+++ b/addon/components/au-link.js
@@ -1,4 +1,5 @@
 import Component from '@glimmer/component';
+import { deprecate } from '@ember/debug';
 
 const SKIN_CLASSES = {
   primary: 'au-c-link',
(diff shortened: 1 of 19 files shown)


addon/components/au-link.js
```diff
@@ -1,6 +1,30 @@
...
 
 export default class AuLink extends Component {
+  get route() {
+    if (this.args.linkRoute) {
+      deprecate(
+        '@linkRoute is deprecated, use @route instead',
+        false,
+        {
+          id: '@appuniversum/ember-appuniversum.au-link.linkRoute-argument',
+          until: "0.7.0",
```
Windvis, line 12: Not sure about these version numbers but it's required by the deprecate util so I just provided something.

Windvis:
The second commit is just replacing the internal usage of `@linkRoute` which touches a lot of files.

PR #116, opened 2021-06-09
Fix some `<AuDatePicker>` issues

### LINKED ISSUE #103: [AuDatePicker] Timezone issue
When initializing the date picker with a date that has a non-UTC timezone, that info will be lost.

[All dates are converted to an ISO 8601 format](https://github.com/appuniversum/ember-appuniversum/blob/52566adf81dc1225bf2573eaa9a1aa5b2dd9b641/addon/components/au-date-picker.js#L131), since that's what `<duet-date-picker>` expects. This is done using the [`Date.prototype.toISOString` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) which means it's converted to the UTC timezone.

From the docs:
> The timezone is always zero UTC offset, as denoted by the suffix "Z".

This will have the side effect that some dates will return the ISO string of the day before (or after when the user is in a negative timezone offset).

For example: 12 may 2021 00:00 in a +2 timezone will be converted to `2021-05-11T22:00:00.000Z` by `toISOString`.

I don't think it's a critical issue since those late hours will probably not be chosen by end-users (and so the wrapping bug will not occur), but it's not correct either way.

The easiest fix for this would be to use the [date-fns formatISO utility](https://date-fns.org/v2.21.3/docs/formatISO). Which does respect the timezone:
image

Windvis:
This fixes 2 issues:
1. The timezone issue as described in #103 
2. A timing issue with the duet-date-picker component: In some cases it was possible for a `<AuDatePicker>` to be rendered before duet-date-picker was registered which made the configuration of the adapter fail. This would result in the input not showing the date in the correct format (since duet would default to its own config). Instead of making the app wait until duet was registered, I opted for a "lazy-loaded" version. Now duet will only be loaded by projects that actually use the `<AuDatePicker>` component.

Closes #103


addon/components/au-date-picker.js
```diff
@@ -1,7 +1,9 @@
 import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 import { action } from "@ember/object";
 import { assert, runInDebug } from '@ember/debug';
 import { guidFor } from '@ember/object/internals';
+import formatISO from 'date-fns/formatISO'
```
Windvis, line 6: FYI I verified that this was properly tree shaken and not all the date-fns were included:

Before this PR:
` - dist/assets/vendor-970d88007936a1cefb44c371d9bd4ca3.js: 2.28 MB (487.25 KB gzipped)`

Now:
` - dist/assets/vendor-5dd17018d26aba4293fbfe3e213a05bf.js: 2.28 MB (488.24 KB gzipped)`

So that seems about right.

Issue #85, opened 2021-03-31
Buttons: conform to latest style Vlaanderen?


Wolfr:
![image](https://user-images.githubusercontent.com/12690/113103832-1383eb80-9200-11eb-99d9-67f84bde12bd.png)

I have noticed that the buttons on Vlaanderen.be are different and more modern: they are rounded and they use a medium font weight.

A quick experiment shows me that we can easily apply this by applying the following code to `.auk-button` (in our own codebase):

```
  font-weight: 500;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
  border-radius: .3rem;
```

This gives following result:

<img src="https://user-images.githubusercontent.com/12690/113103938-31e9e700-9200-11eb-8c30-670b1ed65e45.png" width="300">
<img src="https://user-images.githubusercontent.com/12690/113103957-37dfc800-9200-11eb-8242-7de5376c1b2e.png" width="300">

It seems to me that this is something we could apply.

What do you think?

PR #133, opened 2021-08-12
Rework buttons to conform to Webuniversum

Dietr:
See also #85 

This PR reworks all buttons and related elements to conform to the latest webuniversum style:
- buttons are a lighter blue and have rounded corners
- input fields, cards and alerts take on the same rounded corners
- links have the same tint blue as the buttons

This will require some testing and tweaking to existing interfaces.


mietcls:
Looks good, small steps towards the new webuniversum style. Some thoughts to fix later, and discuss:

- We need a fix for the tertiary buttons and anchor buttons; or we change it everywhere or we provide a back-up. Some buttons break in loket. TBD
- Review hover contrast; it's very subtle.
- Our pills look very similar; but the current webuniversum pills have some accessibility issues. We need to document this in appuniversum.

PR #135, opened 2021-08-23
allow users to specify whether au-card is open on init or not

nvdk:
there's room for improvement here, but it works for the basic case.

### DIFF
--- a/tests/dummy/app/templates/docs/components/au-card.md
+++ b/tests/dummy/app/templates/docs/components/au-card.md
@@ -149,7 +149,7 @@
 
 {{#docs-demo as |demo|}}
   {{#demo.example name='au-card-expandable.hbs'}}
-    <AuCard @expandable={{true}} @flex={{true}} @divided={{true}} as |c|>
+    <AuCard @expandable={{true}} @flex={{true}} @divided={{true} @isOpenInitially={{true}} as |c|>
       <c.header @badgeSkin="brand" @badgeIcon="users">
         <AuHeading @level="2" @skin="4">
           Title for card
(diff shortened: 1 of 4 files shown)


Windvis:
The only remark I have is that using the same argument name as the internal proprty name might be confusing when reading the code again in a few weeks :smile:.

nvdk:
I've renamed the argument to isOpenInitially and added it to do the doc

nvdk:
> Not sure if I understand what it does exactly or what the use case is but doesn't seem to break anything so ok for me. Do we need to add some extra information to the card component documentation page?

the idea is that in some uses cases the card can already be open without the user having to click, so it's still closable by the user if needs be. It's a somewhat limited use case, I mostly need it in GN atm for a not so pretty hack :see_no_evil:

tests/dummy/app/templates/docs/components/au-card.md
```diff
@@ -149,7 +149,7 @@
 
 {{#docs-demo as |demo|}}
   {{#demo.example name='au-card-expandable.hbs'}}
-    <AuCard @expandable={{true}} @flex={{true}} @divided={{true}} as |c|>
+    <AuCard @expandable={{true}} @flex={{true}} @divided={{true} @isisOpenInitially={{true}} as |c|>
```
Windvis, line 152: `isis` typo :grimacing:

tests/dummy/app/templates/docs/components/au-card.md
```diff
@@ -149,7 +149,7 @@
 
 {{#docs-demo as |demo|}}
```
Windvis, line 150: I think it should be added to the argument table at the bottom of the docs page as well?

Windvis (approved):

PR #137, opened 2021-08-27
Add a total row fallback to the data table component

### LINKED ISSUE #136: [Bug] AuDataTable total rows issue
When passing a regular array into the table component the total number of rows isn't displayed correctly. This can be seen in the docs page of that component already: https://appuniversum.github.io/ember-appuniversum/docs/components/au-data-table

`1 - 3 van` 

The reason for this is that the array doesn't have a meta object property which is what the pagination component uses to determine the total.

I think it would be a good idea to use the total amount of records that are currently visible as a fallback for the missing meta property.

Windvis:
This shows the total number of visible rows when the pagination meta object doesn't include a count. This means that arrays that are created client side are supported as well.

Closes #136

### DIFF
--- a/addon/components/au-data-table/number-pagination.hbs
+++ b/addon/components/au-data-table/number-pagination.hbs
@@ -1,6 +1,6 @@
 <div class="au-c-pagination">
   <p>
-    <span class="au-u-hidden-visually">Rij </span><strong>{{startItem}} - {{endItem}}</strong> van {{total}}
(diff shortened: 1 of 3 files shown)


Windvis:
The ember-data-table addon [doesn't show the total if it is undefined](https://github.com/mu-semtech/ember-data-table/blob/13635da7f7c3f2690bb99808c4fe3671c04972ee/addon/templates/components/number-pagination.hbs#L4), so that's a potential solution as well.

Dietr (approved):

PR #140, opened 2021-09-09
Table component

Dietr:
Simple table component.

@Windvis Can you review the use of the named blocks? Not sure if I use this correctly or if we can/should also make components out of the tr, th, td elements.


Windvis:
That seems about right. You could yield components from named blocks if you need that but that doesn't seem to be the case? Or did you use workarounds for something?

If those elements don't need any extra classes or configuration I think it might be better to keep it like this for now?

PR #145, opened 2021-09-16
wip: `AuSelect` component

### LINKED ISSUE #143: Create a simple select component
ember-appuniversum has styles for the very popular ember-power-select addon. This means that most projects use that addon for all select related scenarios even when non of the advanced features are needed. The [addon isn't 100% accessible either](https://github.com/cibernox/ember-power-select/issues/871) and it seems impossible to make it behave like a native select.

Adding an `AuSelect` component which uses the native select element could be a good alternative for the basic select use-cases. This would have the benefit that it is accessible out of the box.

Windvis:
A WIP experiment to create a simple power select alternative. I tried to use the same argument names to make transitioning easier.

This is what the usage looks like at the moment:
```hbs
    <AuSelect
      @placeholder="Kies een provincie"
      @options={{this.provinces}}
      @selected={{this.selectedProvince}}
      @onChange={{fn (mut this.selectedProvince)}} 
      id="au-select-province"
      as |province|
    >
      {{province.label}}
    </AuSelect>
```

Remaining things:
- styling :see_no_evil: 
- [Option grouping](https://ember-power-select.com/docs/groups)
- ~multi-select support?~ [Apparently this isn't a UX-friendly or accessible option.](https://www.24a11y.com/2019/select-your-poison-part-2/)
- `@allowClear` argument
- tests
- probably other things, it's a quick draft at the moment.

Closes #143

### DIFF
--- a/addon/components/au-select.hbs
+++ b/addon/components/au-select.hbs
@@ -1,0 +1,15 @@
+<select {{on "change" this.handleSelectionChange}} ...attributes>
+  {{#if @placeholder}}
+    {{!
+      TODO this probably needs to be influenced by the `@allowClear` argument as well.
+      If that is false this first option should not be visible, or disabled if the placeholder is set.
+    }}
+    <option value={{this.CLEAR_VALUE}}>{{@placeholder}}</option>
+  {{/if}}
+  {{#each @options as |optionValue index|}}
+    {{!-- TODO: add option grouping support --}}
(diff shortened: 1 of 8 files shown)


Windvis:
Closing as I haven't touched it in years and projects don't seem to have a requirement for it either. We can reopen / be inspired by this one once we do have this requirement.

PR #154, opened 2021-11-05
Update to the Ember cli 3.28 blueprint

### LINKED ISSUE #127: Ember Globals deprecation messages
Updating an app that uses ember-appuniversum to Ember 3.27 triggers the following warning:
```
DEPRECATION: Usage of the Ember Global is deprecated. You should import the Ember module or the specific API instead.

See https://deprecations.emberjs.com/v3.x/#toc_ember-global for details.

Usages of the Ember Global may be caused by an outdated ember-cli-babel dependency. The following steps may help:

* Upgrade the following addons to the latest version:
  * @appuniversum/ember-appuniversum
  * ...

### Important ###

In order to avoid repeatedly showing the same deprecation messages, no further deprecation messages will be shown for usages of the Ember Global until ember-cli-babel is upgraded to v7.26.6 or above.

To see all instances of this deprecation message, set the `EMBER_GLOBAL_DEPRECATIONS` environment variable to "all", e.g. `EMBER_GLOBAL_DEPRECATIONS=all ember test`.

### Details ###

Prior to v7.26.6, ember-cli-babel sometimes transpiled imports into the equivalent Ember Global API, potentially triggering this deprecation message indirectly, even when you did not observe these deprecated usages in your code.

The following outdated versions are found in your project:

* ember-cli-babel@6.18.0, currently used by:
  * ...
  * ember-truth-helpers@2.1.0
    * Added by @appuniversum/ember-appuniversum@0.4.8 > ember-data-table@2.0.1

Note: Addons marked as "Dormant" does not appear to have any JavaScript files. Therefore, even if they are using an old version ember-cli-babel, they are unlikely to be the culprit of this deprecation and can likely be ignored.
 [deprecation id: ember-global] See https://deprecations.emberjs.com/v3.x/#toc_ember-global for more details.
    registerHandler/HANDLERS[type]@http://localhost:4200/assets/vendor.js:36640:17
    raiseOnDeprecation@http://localhost:4200/assets/vendor.js:36534:13
    registerHandler/HANDLERS[type]@http://localhost:4200/assets/vendor.js:36640:17
    invoke@http://localhost:4200/assets/vendor.js:36652:23
    deprecate@http://localhost:4200/assets/vendor.js:36608:28
    get@http://localhost:4200/assets/vendor.js:477:53
    @http://localhost:4200/assets/vendor.js:69473:7
    @http://localhost:4200/assets/vendor.js:69498:3
```

It sounds like updating the blueprint of the addon should fix it.

Windvis:
This triggers _a lot_ of deprecations and linting errors but nothing seems broken.

A lot of deprecations are triggered by ember-addon-docs but updating its version is tricky since we override its templates (which also trigger deprecations :see_no_evil:).

Closes #127


Dietr:
Does this break support for apps running an older ember-cli?
Regarding the ember-addon-docs. We're thinking of moving the docs to storybooks so this could also fix the deprecations in the future.

Windvis:
> Does this break support for apps running an older ember-cli?

No, the Ember version is only used in the dummy app. As long as we don't use any code that isn't compatible with older versions nothing changes for consuming applications. The main benefit is that it will allow us to notice deprecation warnings in code that ships to consumers (which would also cause deprecations there).

> Regarding the ember-addon-docs. We're thinking of moving the docs to storybooks so this could also fix the deprecations in the future.

:+1:

PR #178, opened 2022-01-24
Enable prettier and run `npm run lint:fix`

### LINKED ISSUE #174: Enable Prettier
I would enable Prettier again since it takes away the burden of formatting code manually and makes it more consistent as an extra bonus. Simply running `npm run lint:fix` should fix all the Prettier related issues. It will be a lot of changes but it's best to bundle those in 1 commit I think.

Windvis:
This enables prettier and adds configuration for template files so those can be formatted (manually) as well.

Closes #174


.prettierrc.js
```diff
@@ -2,4 +2,12 @@
 
 module.exports = {
   singleQuote: true,
+  overrides: [
+    {
+      files: '*.hbs',
+      options: {
+        singleQuote: false,
```
Windvis, line 9: This allows us to use prettier for handlebars files without it changing all quotes to `'`.

PR #188, opened 2022-02-10
Resolving Ember 3.28 deprecations

benjay10:
This PR is for tracking and resolving most Ember version 3.28 deprecation warnings.


benjay10:
**Fixes only cover the code paths that are taken by the test scenario's!**

As of now, most fixes involved adding `this.` on accessors in components and changing `hasBlock` to `(has-block)`. However, there were no deprecation warnings for the `hasBlock`, I just happen to know that this needs to be changed.

Next up is to explore the remaining code paths, and to rewrite component constructions (`{{component ...}}` into `<Component ...>`) for consolidation.

addon/components/au-data-table/content.hbs
```diff
@@ -1,14 +1,14 @@
 {{! template-lint-disable table-groups }}
 <div class="au-c-data-table__wrapper">
-  <table class="au-c-data-table__table {{tableClass}}" ...attributes>
-    {{#if hasBlock}}
+  <table class="au-c-data-table__table {{this.tableClass}}" ...attributes>
+    {{#if (has-block)}}
       {{yield (hash
-          header=(component "au-data-table-content-header" enableSelection=enableSelection enableLineNumbers=enableLineNumbers data-table=data-table)
-          body=(component "au-data-table-content-body" content=content enableSelection=enableSelection enableLineNumbers=enableLineNumbers noDataMessage=noDataMessage onClickRow=(optional onClickRow) data-table=data-table)
+          header=(component "au-data-table-content-header" content=this.content enableSelection=this.enableSelection enableLineNumbers=this.enableLineNumbers data-table=this.data-table)
```
Windvis, line 6: Why does the `content` value need to be passed in? It seems it wasn't needed previously?

addon/components/au-data-table/content-body.hbs
```diff
@@ -1,25 +1,26 @@
...
-    {{#each wrappedItems as |wrapper index|}}
-      <tr class={{if (includes wrapper.item data-table.selection) "selected"}} {{action (optional onClickRow) wrapper.item}}>
-        {{#if enableSelection}}
+  {{#if this.content}}
+    {{#each this.wrappedItems as |wrapper index|}}
+      <tr class={{if (includes wrapper.item this.data-table.selection) "selected"}} {{action (optional this.onClickRow) wrapper.item}}>
+        {{#if this.enableSelection}}
           <td class="is-selectable">
-            {{input type="checkbox" checked=wrapper.isSelected click=(action "updateSelection" wrapper)}}
+            {{!--{{input type="checkbox" checked=wrapper.isSelected click=(action "updateSelection" wrapper)}}--}}
```
Windvis, line 9:
```suggestion
```

Windvis (changes requested):
Thanks for working on this!

Windvis:
> and to rewrite component constructions ({{component ...}} into <Component ...>) for consolidation.

I wouldn't spend too much time refactoring the data table components and only solve the deprecations. Ideally we move away from the template overriding we do at the moment so that means most of these components will be removed anyway.

addon/components/au-data-table/content.hbs
benjay10: Luckily no harm, but total miss of mine.

addon/components/au-data-table/text-search.hbs
```diff
@@ -1,5 +1,5 @@
 <div class="au-c-data-table__search">
-  {{input value=value placeholder=placeholder class="au-c-input au-c-input--block"}}
+  <Input value={{this.value}} placeholder={{this.placeholder}} class="au-c-input au-c-input--block" />
```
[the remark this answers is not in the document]
Windvis: > Is using @ because of 2-way binding only? The placeholder is using this. and it's getting the correct values.

Yes, at the moment typing in the input field will not update `this.value`. It would be fine if there was an action attached to the input event that does the updating but that isn't the case here. So I assume this depended on the 2-way binding to work.

> I don't see this component constructed with a `@value` argument, so where does this value come from anyway?

It extends the component from the ember-data-table addon which [sets the value on init (and in some other places)](https://github.com/mu-semtech/ember-data-table/blob/eb1ec672c34b5131ba4048f60670d743606cb61b/addon/components/text-search.js#L17). I haven't used the component myself so I'm not sure what issue you mean.

addon/components/au-data-table/text-search.hbs
```diff
@@ -1,5 +1,5 @@
 <div class="au-c-data-table__search">
-  {{input value=value placeholder=placeholder class="au-c-input au-c-input--block"}}
+  <Input value={{@value}} placeholder={{this.placeholder}} class="au-c-input au-c-input--block" />
```
Windvis, line 2: Sorry, I might have been unclear with what I meant. It should be passed as an argument, not an attribute for the 2-way binding to work. It's the `Input` component that does the 2-way binding magic.
```suggestion
  <Input @value={{this.value}} placeholder={{this.placeholder}} class="au-c-input au-c-input--block" />
```

`@value` only returns the correct value here because in `@ember/components` there is no difference between component arguments and class properties.

addon/components/au-data-table/text-search.hbs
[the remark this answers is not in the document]
benjay10: Okay, nevermind. Small misunderstanding on my part. Using `@value` (in the correct place) indeed solves all of my problems.

Dietr (approved):

benjay10:
This PR might not be complete, but it would be nice to see these smaller fixes finally out. Nothing should be a breaking change.

Windvis:
@benjay10 I merged it, but I don't have the needed npm permissions to create a new release.

benjay10:
Thanks @Windvis! Me either, but @Dietr knows about this PR, so it's going to be OK.

PR #192, opened 2022-02-15
`AuInput` inputmask improvements

Windvis:
This updates to the newest version of ember-inputmask5 addon which lazy-loads the inputmask library.This means apps will only ship the code to the user if the mask functionality is actually used.

Build sizes:
```
   Before update
 - dist/assets/vendor-93db5da83a6121acea322bc0962a17e1.js: 1.24 MB (282.9 KB gzipped)

   After update 2 bundles and the chunk will be lazy-loaded
 - dist/assets/vendor-86b0ce0f7ca457fbbc62d6acbe3b1c88.js: 1.14 MB (253.65 KB gzipped)
 - dist/assets/chunk.550.8e41283faa149689417f.js: 98.55 KB (29.57 KB gzipped)
```

I also added an `onChange` event to the input component which can be used to retrieve the unmasked value.

It might be a good idea to deprecate the 2-way-binding and let users also pass in a change handler to update the value even if they're not using the mask feature. This would make the component more consistent and allows us to switch to a regular input element.


Windvis:
I put it in draft since it doesn't seem to lazy-load inputmask in an actual app.

Dietr (approved):

Issue #194, opened 2022-02-15
New scss files that group imports


Windvis:
The current docs lists 2 options for importing the scss:

1. a single import which imports everything: `@import "ember-appuniversum";` which has the benefit that the addon decides what goes in the build, but apps don't have control over the order.
2. List all the imports in the app. This provides the most control but requires manual maintenance of the imports when updating

There is a middle ground between these 2 options that might be worth it. We could create scss files for each "type" of import. Projects can then import these files if they need more control than option 1, but still want to import all the styles from appuniversum.

This is what it could look like if you create a file for each "title" in the huge import list, but there might be better ways to group things.
```scss
@import "ember-appuniversum/variables";
@import "ember-appuniversum/mixins";
@import "ember-appuniversum/generic";
@import "ember-appuniversum/elements";
@import "ember-appuniversum/objects";
@import "ember-appuniversum/components";
@import "ember-appuniversum/plugins";
@import "ember-appuniversum/utilities";
```

Apps can then add new app specific styles in the correct place.

Windvis:
The loket app would benefit from this since it still depends [on other styles and components](https://github.com/lblod/frontend-loket/blob/development/app/styles/app.scss#L56-L61), and replacing the appuniversum styles with a single import changes the order of the classes which might affect the app and issues would be hard to detect.

Dietr:
Good idea, i did something similar for the rdfa-editor and this could also improve that implementation. Seems pretty easy to implement.

PR #197, opened 2022-02-16
Group scss imports

Dietr:
See #194

### DIFF
--- a/app/styles/ember-appuniversum.scss
+++ b/app/styles/ember-appuniversum.scss
@@ -15,100 +15,16 @@
  * ELEMENTS: Plain html tag styling
  * OBJECTS: OOCSS unstyled objects (namespace: '.au-o-...')
  * COMPONENTS: Components (namespace: '.au-c-...')
+ * PLUGINS: Third-party styles
  * UTILITIES: Single function helper classes (namespace: '.au-u-...')
  */
 
 
-// VARIABLES
-@import "appuniversum/s-colors.scss";
-@import "appuniversum/s-global";
-@import "appuniversum/s-utilities";
-@import "appuniversum/s-root";
-
-
-// MIXINS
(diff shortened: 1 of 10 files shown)


app/styles/ember-appuniversum.scss
```diff
@@ -15,100 +15,16 @@
...
-@import "appuniversum/u-font-family";
-@import "appuniversum/u-font-weights";
-@import "appuniversum/u-max-widths";
-@import "appuniversum/u-paragraphs";
-@import "appuniversum/u-print";
-@import "appuniversum/u-responsive-spacings";
-@import "appuniversum/u-spacings";
-@import "appuniversum/u-visible";
-@import "appuniversum/u-widths";
+@import "ember-appuniversum/a-settings";
```
Windvis, line 23: Silly question, but what system are we using for the file naming? What does the `a-` prefix mean here?

stories/1-outline/Getting-started.stories.mdx
```diff
@@ -72,8 +72,19 @@ Then load the SCSS in your `app.scss` file:
 @import "ember-appuniversum";
 ```
 
+If you want more control over the order that SCSS partials are rendered you can import them as groups.
+``` js
```
Windvis, line 76:
```suggestion
``` scss
```

Windvis (approved):
Nice! Seems to work just as expected when I tried it in loket. :+1:

app/styles/ember-appuniversum.scss
Dietr: It's based on ITCSS (https://appuniversum.github.io/appuniversum/guidelines/architecture/). The reason of the prefixes is to easily spot files and their purpose (also having them grouped alphabetically is quite handy). I used the "a" here for "all". "g" (group) was already taken by generic.

PR #209, opened 2022-02-23
Add an option to localize the `AuDatePicker` component in Dutch

Windvis:
This adds a new config flag that replaces the English localization with a Dutch version when enabled. This option is disabled by default to keep backwards-compatibility, but it will become the default in the next major release since that's also the default language in all apps that use appuniversum. For this reason we also show a deprecation message so users know they need to manually switch to the Dutch version.

The Dutch version can be enabled by adding the following snippet to the app's `ember-cli-build.js` file:
```js
  let app = new EmberApp(defaults, {
    //.. other options
    '@embroider/macros': {
      setConfig: {
        '@appuniversum/ember-appuniversum': {
          dutchDatePickerLocalization: true,
        },
      },
    },
```


Windvis:
It might be nice to document this in storybook somewhere, but I'm not sure how that works. Any pointers?

addon/components/au-date-picker.js
```diff
@@ -1,72 +1,66 @@
...
-  ],
+  dayNames: getLocalizedDays(),
+  monthNames: getLocalizedMonths(),
+  monthNamesShort: getLocalizedMonths('short'),
 };
 
+if (macroCondition(getOwnConfig().dutchDatePickerLocalization)) {
+  Object.assign(DEFAULT_LOCALIZATION, {
+    buttonLabel: 'Kies een datum',
+    placeholder: 'DD.MM.JJJJ',
```
Windvis, line 39: I've used `DD.MM.JJJJ` for now since it seems the most official, but it can easily be updated when ABB decides on the format we should use across projects.

index.js
```diff
@@ -19,6 +24,17 @@ module.exports = {
...
       'node_modules/@appuniversum/appuniversum'
     );
+
+    let addonOptions = app.options[this.name];
+    if (addonOptions && addonOptions.dutchDatePickerLocalization) {
+      this.options[
+        '@embroider/macros'
+      ].setOwnConfig.dutchDatePickerLocalization = true;
+    } else {
+      this.ui.writeDeprecateLine(
```
Windvis, line 34: This will show a warning in terminal when building the app since it's possible to miss the browser deprecation. Is that overkill?

Windvis:
Draft since the date format hasn't been decided yet.

index.js
Windvis: I've removed it again, not all projects might be using the component so it would only be unnecessary noise.

addon/components/au-date-picker.js
```diff
@@ -76,6 +62,32 @@ export default class AuDatePickerComponent extends Component {
...
+      \`\`\`
+      // ember-cli-build.js
+      '@appuniversum/ember-appuniversum': {
+        dutchDatePickerLocalization: true,
+      },
+      \`\`\``,
+      isNewImplementation,
+      {
+        id: '@appuniversum/ember-appuniversum.au-date-picker.english-localization',
+        until: '3.0.0',
```
Windvis, line 84: Since the v2 release is close I opted for v3 instead.

package.json
```diff
@@ -50,6 +50,7 @@
   ],
   "dependencies": {
     "@duetds/date-picker": "^1.4.0",
+    "@embroider/macros": "^1.2.0",
```
Windvis, line 53: I can probably bump this to the latest version.

addon/utils/date.js
```diff
@@ -0,0 +1,6 @@
+export function formatDate(date) {
+  let day = `${date.getDate()}`.padStart(2, '0');
+  let month = `${date.getMonth() + 1}`.padStart(2, '0');
```
Windvis, line 3: I need to double check if it's `01-01-2022` or `1-1-2022` 😬.

Issue #218, opened 2022-03-24
`.au-c-content--tiny` font size


Windvis:
The idea I had for the `.au-c-content--tiny` modifier was that it matched the style of help text component so we could add generic content to it and have it styled consistently. 

It seems the font size is a tiny bit larger than the help-text text so it looks a bit weird imo:
![image](https://user-images.githubusercontent.com/3533236/159929861-b2003025-1bdf-469a-9f6c-b13073be915c.png)

Compared to the custom version we used in loket:
![image](https://user-images.githubusercontent.com/3533236/159929983-67f4e16f-a1a7-4dbc-b798-5d894b42d1af.png)

Any objections to make the font sizes match?

Windvis:
Fixed by #225

Issue #210, opened 2022-02-24
Radio button grouping


Windvis:
I've seen some projects use a layout similar to something like this for a group of radio buttons:

```hbs
<AuLabel for="radios">Some label that describes the radio options</AuLabel>
<ul id="radios">
  <li><AuControlRadio @label="option 1" /></li>
  <li><AuControlRadio @label="option 2" /></li>
...
</ul>
```

I think the more semantically correct version would use a [`fieldset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) and `legend` combination instead?

```hbs
<fieldset>
  <legend>Some label that describes the radio options</legend>
  <AuControlRadio @label="option 1" />
  <AuControlRadio @label="option 2" />
...
</fieldset>
```

At the moment appuniversum doesn't ship any styles for this setup I think. Would it make sense to add some? 

Or a dedicated component that could provide some abstraction for it?

```hbs
<AuRadioGroup @error={{this.hasError}} @required={{this.isRequired}} @disabled={{this.isDisabled}}>
  <:label>Some label that describes the radio options</label>
  <:radios as |Radio|>
    <Radio @label="option 1" />
    <Radio @label="option 2" />
  </radios>
</AuRadioGroup>

Windvis:
A more simple alternative could be to allow changing the rendered element for the `AuLabel` component, so that it can also be used for the legend of a fieldset. The wrapper component doesn't provide a huge benefit over the native fieldset, so maybe just documenting that layout is better?

Dietr:
Adding a legend and fieldset component is an option. The only reason is so we can provide a consistent style without having to add classes on it.

I think it's best to avoid making components like AuLabel behave like a legend. Some components already have that mistake.

Windvis:
Fixed by #225

Issue #191, opened 2022-02-15
Form layout components


Dietr:
More options for forms are needed:
- label with pre option to create a grouped label-form layout
- form row component for better form layouts: a way to quickly layout forms

PR #225, opened 2022-04-01
Add form patterns and missing form components

Dietr:
- [x] Add pre label option and form row component  #191
- [x] Add radio label grouping #210
- [x] Improve helptext so it can handle longer content #218
- [x] Provide examples of form element usage in context


addon/components/au-fieldset/legend.hbs
```diff
@@ -0,0 +1,28 @@
...
+    <AuBadge @icon="alert-triangle" @size="small" @skin="warning" class="au-u-margin-right-tiny" />
+    {{/if}}
+    {{#if this.error}}
+    <AuBadge @icon="cross" @size="small" @skin="error" class="au-u-margin-right-tiny" />
+    {{/if}}
+    {{yield "legend"}}
+    {{#if @required}}
+      {{#if @inline}}
+        {{#if @requiredLabel}}
+        <span class="au-c-fieldset__legend-addendum">{{@requiredLabel}}</span>
```
Windvis, line 14: nitpick, I think it would be "better" to use `au-c-fieldset-legend` and `au-c-fieldset-legend__addendum` instead?

addon/components/au-fieldset/legend.js
```diff
@@ -0,0 +1,29 @@
...
+
+export default class AuFieldsetLegend extends Component {
+  get skin() {
+    if (this.args.skin == '1') return 'au-u-h1';
+    if (this.args.skin == '2') return 'au-u-h2';
+    if (this.args.skin == '3') return 'au-u-h3';
+    if (this.args.skin == '4') return 'au-u-h4';
+    if (this.args.skin == '5') return 'au-u-h5';
+    if (this.args.skin == '6') return 'au-u-h6';
+    if (this.args.skin == 'functional') return 'au-u-h-functional';
```
Windvis, line 11: What's this for exactly?

addon/components/au-fieldset/legend.hbs
```diff
@@ -0,0 +1,28 @@
+{{#if (has-block)}}
+<div ...attributes>
```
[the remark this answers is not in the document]
Dietr: The div is there to allow the inline alignment option. A legend doesn't accept flexbox positioning so this is a workaround to enable that.

addon/components/au-fieldset/legend.js
Dietr: So you can change the font size of the legend when using this as a fieldset grouping multiple elements. I added an example for this.

stories/4-components/Forms/AuFieldset.stories.js
```diff
@@ -0,0 +1,75 @@
...
+        @inline={{this.inline}}
+        @required={{this.required}}
+        @requiredLabel={{this.requiredLabel}}
+        @error={{this.error}}
+        @warning={{this.warning}}
+      >
+        Legend
+      </f.legend>
+      <f.content>
+        <AuInput id="input" />
```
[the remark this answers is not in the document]
Dietr: Yes, I added the radio list and group inputs example.

addon/components/au-fieldset/legend.js
Windvis: Yea, but specifically the `functional` version?

addon/components/au-fieldset.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@glimmer/component';
+
+export default class AuFieldset extends Component {
+  get alignment() {
+    if (this.args.alignment == 'inline') return 'au-c-fieldset--inline';
+    if (this.args.alignment == 'pre')
+      return 'au-c-fieldset--inline au-c-fieldset--pre';
+    if (this.args.alignment == 'post')
```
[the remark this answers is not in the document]
Dietr: There's some inputs with pre alignment in RDFA editor plugins. The post option is here for consistency with the form-row component. Might indeed be unnecessary as I've never used the post option.

addon/components/au-fieldset/legend.js
Dietr: That's a title size used in Kaleidos and webuniversum. I want to provide the same sizing options as a title here.

Windvis: Ok :+1: Hadn't encountered it before.

Windvis: Would it make sense to also add a `label` skin? The cases I've seen would just want it to look like a label instead of a title.

addon/components/au-fieldset.js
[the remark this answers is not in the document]
Windvis: But those wouldn't be `<legends>`, right? I don't see a use-case where a legend would be used with these alignment options. It feels like we're making the component more complex than it should be. I would rather only add features if actually needed. So unless you're sure there are valid use cases for it I would drop that functionality here for now.

addon/components/au-fieldset/legend.js
Dietr: The default size and styling is the same as a label or (au-u-h6) so not sure if we need it as a seperate option.

PR #229, opened 2022-04-11
Add a `@type` argument to the AuInput component

### LINKED ISSUE #217: Passing a `type` attribute to the `AuInput` component no longer works
It seems that passing a type attribute to the `AuInput` component no longer works on newer Ember versions. I would guess 3.27+ since that included a refactor of all the built-in components.

I think this only worked by accident before since it's [clearly documented that we should pass it as an argument](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input#input-html-attributes-to-avoid) instead `@type="number"`.

The problem is that we don't pass that through to the underlying Input components so that's not possible yet if we want to use `AuInput`.

Windvis:
Ember's Input component requires that the type is passed as an argument instead of an attribute. [This has always been documented like this](https://api.emberjs.com/ember/release/classes/Ember.Templates.components/methods/Input#input-html-attributes-to-avoid) but somewhere between 3.24 and 3.28 it stopped supporting the attribute version.

Closes #217

### DIFF
--- a/addon/components/au-input.hbs
+++ b/addon/components/au-input.hbs
@@ -1,20 +1,66 @@
 {{#if @icon}}
   <span class="au-c-input-wrapper {{this.iconAlignment}} {{this.width}}">
     {{#if @mask}}
-    <Inputmask class="au-c-input au-c-input--mask {{this.error}} {{this.warning}} {{this.width}} {{this.disabled}}" @value={{@value}} @mask={{@mask}} @placeholder={{@maskPlaceholder}} @update={{this.handleChange}} disabled={{@disabled}} ...attributes/>
+      <Inputmask
+        @value={{@value}}
+        @mask={{@mask}}
+        @placeholder={{@maskPlaceholder}}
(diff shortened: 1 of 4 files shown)


addon/components/au-input.hbs
```diff
@@ -1,20 +1,66 @@
 {{#if @icon}}
   <span class="au-c-input-wrapper {{this.iconAlignment}} {{this.width}}">
     {{#if @mask}}
-    <Inputmask class="au-c-input au-c-input--mask {{this.error}} {{this.warning}} {{this.width}} {{this.disabled}}" @value={{@value}} @mask={{@mask}} @placeholder={{@maskPlaceholder}} @update={{this.handleChange}} disabled={{@disabled}} ...attributes/>
+      <Inputmask
```
Windvis, line 4: The lines were getting very long so I formatted the template with Prettier.

Dietr (approved):

PR #231, opened 2022-04-14
Use a native input element in the `AuControlCheckbox` component

Windvis:
The `Input` component provides no benefits since we already required the `onChange` argument. The 2-way-binding isn't needed here.

### DIFF
--- a/addon/components/au-control-checkbox.hbs
+++ b/addon/components/au-control-checkbox.hbs
@@ -1,5 +1,17 @@
-<label for="{{@identifier}}" class="au-c-control au-c-control--checkbox {{this.disabled}}">
-  <Input class="au-c-control__input" id="{{@identifier}}" name="{{@name}}" disabled={{@disabled}} @type="checkbox" @checked={{@checked}} {{on "change" this.onChange}} ...attributes />
+<label
+  for={{@identifier}}
+  class="au-c-control au-c-control--checkbox {{this.disabled}}"
+>
+  <input
+    class="au-c-control__input"
+    id={{@identifier}}
+    name={{@name}}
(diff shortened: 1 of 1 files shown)


addon/components/au-control-checkbox.hbs
```diff
@@ -1,5 +1,17 @@
-<label for="{{@identifier}}" class="au-c-control au-c-control--checkbox {{this.disabled}}">
-  <Input class="au-c-control__input" id="{{@identifier}}" name="{{@name}}" disabled={{@disabled}} @type="checkbox" @checked={{@checked}} {{on "change" this.onChange}} ...attributes />
+<label
+  for={{@identifier}}
```
Windvis, line 2: Technically this isn't needed since the input is nested in the label.

Windvis:
> I'm not sure if we should consider this a breaking change. Theoretically people could only provide the @checked argument and depend on the 2-way binding, but it was never documented that way so I consider the Input component an implementation detail that we can swap out. I did a code search in the LBLOD org and couldn't find cases where the 2-way binding was used. 

I asked around in the Discord and since we couldn't really find any cases were the missing 2-way-binding would be problematic, this can be considered a bugfix.

Windvis:
> I'm not sure if we should consider this a breaking change. Theoretically people could only provide the @checked argument and depend on the 2-way binding, but it was never documented that way so I consider the Input component an implementation detail that we can swap out. I did a code search in the LBLOD org and couldn't find cases where the 2-way binding was used. 

I asked around in the Discord and since we couldn't really find any cases were the missing 2-way-binding would be problematic, this can be considered a bugfix.

Issue #223, opened 2022-03-31
Deprecation from ember-modifier


claire-lovisa:
In OP I've started to see some deprecations warnings like shown below that seem to get triggered by the `AuDropdown` component :
```DEPRECATION: ember-modifier (for onClickInside ... ```

After a little check in the app on `ember-modifier`, it seems ember-appuniversums still uses v2.1.2. Maybe a bump would fix the warning ?

```
├─┬ @appuniversum/ember-appuniversum@1.0.3
│ ├─┬ ember-inputmask5@3.2.0
│ │ └── ember-modifier@3.2.1 deduped
│ ├── ember-modifier@2.1.2
│ └─┬ ember-prop-modifier@1.0.1
│   └── ember-modifier@2.1.2
```

Thanks ! :)

Windvis:
I think it's more complicated than simply bumping the version. It seems the latest release of ember-modifier triggers custom deprecations which need actual changes: https://github.com/ember-modifier/ember-modifier/releases/tag/v3.2.0

All our modifier subdependencies need to be updated and our own `on-click-inside` modifier also needs that change. I'll try to look into it ASAP.

Windvis:
These are the addons that need to be updated:
- https://github.com/zestia/ember-auto-focus
- https://github.com/pzuraq/ember-prop-modifier
- https://github.com/sinankeskin/ember-inputmask

Windvis:
The most common warning should be fixed by #234 which is part of the [1.0.6 release](https://github.com/appuniversum/ember-appuniversum/blob/master/CHANGELOG.md#v106-2022-04-15).

Windvis:
We should probably drop both ember-prop-modifier and ember-inputmask and replace them with custom modifiers for our use-case.

ember-prop-modifier isn't maintained and ember-inputmask doesn't really work all that great if you want to do more complex things. They're also thinking of droping the lazy-loading again and we shouldn't ship inputmask as part of the main app bundle, especially if apps don't use the masking functionality.

PR #233, opened 2022-04-15
`on-click-inside` modifier improvements

Windvis:
- add some tests
- resolve the ember-modifier v3.2 deprecation warning (#223)
- simplify the implementation


Windvis:
Heh, I just realized the simplified version is exactly the same as using `{{on "click" this.callback}}` so we don't even need the custom modifier.

PR #237, opened 2022-04-19
Deprecate the `alert` prefixed argument names of the `AuAlert` component

### LINKED ISSUE #236: remove the `alert` prefix from the`AuAlert` component's argument names
It seems that some of the argument names are prefixed with `alert`. I would add new arguments without the prefix and deprecate the prefixed ones, similar to how we did it for other components:

- alertSkin => skin
- alertSize => size
- alertIcon => icon
- alertTitle => title

Windvis:
This deprecates the `alert` prefixed argument names for the `AuAlert` component.

I also added some tests since it didn't have any yet.

Closes #236

### DIFF
--- a/addon/components/au-alert.js
+++ b/addon/components/au-alert.js
@@ -1,29 +1,90 @@
 import Component from '@glimmer/component';
+import { deprecate } from '@ember/debug';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 
 export default class AuAlertComponent extends Component {
-  get alertSkin() {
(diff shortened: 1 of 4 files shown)


Dietr (approved):

addon/components/au-alert.js
```diff
@@ -1,29 +1,90 @@
...
+        },
+      }
+    );
+    return this.args.alertIcon || this.args.icon;
+  }
+
+  get title() {
+    deprecate(
+      '@alertTitle is deprecated, use @title instead',
+      !this.args.alertIcon,
```
Windvis, line 71: Woops, bad copy pasta

Issue #250, opened 2022-05-10
Deprecate the `Input` component usage in `AuInput`


Windvis:
`AuInput` is one of the last components that still uses `Input`'s 2-way-binding internally. This has the downside that it's not suitable for all use-cases and why some apps have to resort to using native inputs with css classes instead.

It would be better to stop using `Input` and let consumers update the value with a change handler. The preference is to do this with the `{{on}}` modifier instead of a custom `@onChange` action since you have more flexibility over the used event (with the downside that the value needs to be read from the event object).

Since this will have a major impact on all apps we can't expect them to change it all at once so a build-time only flag isn't a good option here. We do need a way to detect when the user wants the native element instead of the 2-way-bound `Input`.

Some options:
- add a `@useNativeElement` argument (or similar) which we can use to swap out the implementation.
- deprecate `@value` and push users to the attribute instead? in that case it makes sense to also deprecate some of the other `Input` arguments (`@type`, `@id`) and let users pass these in as attributes as well.
- Deprecate the whole `AuInput` component and create a new `AuNativeInput` component instead

nvdk:
might make (more) sense to provide a separate AuNativeInput component and deprecate the AuInput component?

Windvis:
@nvdk It's an option, but it will have the same end-result as the other options but with a longer name 😄. Any reason why you prefer that route?

Windvis:
> I think it'll be easier to spot what type of input you're using in the app itself. I image a lot of apps will have mixed usage for a while.

Yes, that's true, but the `@useNativeElement` option provides similar benefits I think? It should also be relatively simple to resolve the deprecations so you could switch to the new setup fairly fast I think. 

Anyway, I prefer to keep the `AuInput` name since it's shorter and is the most intuitive name for such a component but I'm not opposed to creating a new one if that's the consensus. I'll add it to the list 👍

abeforgit:
Hit this again today, just to notify that it's still a desired change

Windvis:
@abeforgit Any preference on one of the implementation options? 😄

abeforgit:
agree that AuNativeInput is a clumsy name for what should be the only version of the component
I like breaking stuff so I'd probably just release as a new major and call it a day
but if we're being nice, I suppose deprecating the arguments makes the most sense? The extra argument means that consumers that "do the right thing" immediately will be punished by needing another refactor for when that argument gets removed again

but its a small preference

Windvis:
[...]

PR #252, opened 2022-05-11
AuTimepicker remake

benjay10:
After still experiencing issues after some fixes in https://github.com/appuniversum/ember-appuniversum/pull/249, this PR features a more complete review of the component's code, trying to simplify it and to make it more useful for a specific LBLOD use case.

**This PR includes:**

- Component uses `tracked-toolbox` to keep local copy of tracked parameters, which allows it to update when a parent component changes the time values, while still being able to maintain local state.
- The validation mechanism revised and used more optimally. (No duplicated validation code.)
- Listening on `input` event instead of `focusout` for immediate validation, and to make sure validation also happens on pasting values.
- Listening on `keyup` with more readable code using `event.key` instead of deprecated `event.keyCode`. Now, only ArrowKeyUp and ArrowKeyDown are intercepted, and nothing else.
- Better formatting of the time values, with new formatting function.
- Integration tests for the component


addon/components/au-time-picker.js
```diff
@@ -28,103 +47,44 @@ export default class AuTimePickerComponent extends Component {
...
-    }
-
-    if (elem != 'hourValue') {
-      ++this[elem];
-
-      if (this[elem] >= 59) {
-        this[elem] = 59;
-      }
-    }
+    this[elem] = this.validateTimeValue(this[elem] + 1, elem);
```
Windvis, line 52: While testing out the component in Storybook I noticed that something funky was going on when incrementing a value when the original value was passed in. If for example I set "hours" to `11` in Storybook and increment it, the value jumps to 23. This is because Storybook passes it in as a string, so `this[elem] + 1` will result in `111`, which will set it to the max value after that by the validateTimeValue method.

I think this partly a Storybook configuration issue. It should pass the values as integers, but maybe the arguments need to be validated as well in case consumers do something similar? Maybe an assert should be thrown if the initial argument values are strings? That will probably catch most "misuses".

addon/components/au-time-picker.js
```diff
@@ -1,16 +1,35 @@
...
-      ? 0
-      : this.args.seconds || 0; // ignore 'seconds' argument if 'showSeconds' is false
-  @tracked keyCodes = [8, 9, 13, 33, 34, 37, 39, 46];
+  @localCopy('args.hours') hourValue;
+  @localCopy('args.minutes') minuteValue;
+  @localCopy('args.seconds') secondValue;
+
+  constructor() {
+    super(...arguments);
+    this.hourValue = this.hourValue || 12;
```
Windvis, line 14: localCopy also supports initializers so you can set a default value that way I think? https://github.com/tracked-tools/tracked-toolbox#localcopy

Windvis: Also, I think we want to add some validation here since at the moment it's possible to pass in values outside of the allowed range. Passing in `30` to the `@hours` arg would display that value in the input field that way.

Maybe we can add asserts here in case arg values aren't strings, or when they are outside the range? That will probably catch most cases. It wouldn't detect invalid input when the arg values are updated but that might be fine for now?

addon/components/au-time-picker.js
```diff
@@ -28,103 +47,44 @@ export default class AuTimePickerComponent extends Component {
...
     this.callBackParent(this.getTimeObject);
   }
 
-  /*
-   * calls the function assigned to @onChange by the user with the timeObject as argument
-   */
+  validateTimeValue(value, type) {
+    const max = type === 'hourValue' ? 23 : 59;
+    value = value < 0 ? 0 : value;
+    value = value > max ? max : value;
```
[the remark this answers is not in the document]
Windvis: Since we use the `keyup` event this might not be super useful and the `seconds` input also makes that less ideal than I had in mind I think.

Windvis (commented):
Some remarks but overall a very nice improvement I think!  I would add some tests for this though, since the component doesn't seem to have any yet.

addon/components/au-time-picker.js
Windvis: It seems the Storybook side can be resolved by changing the control type to `number` for the `hours`, `minutes` and `seconds` args and setting the default value to a number instead of a string. That seemed to work here at least. The Storybook docs aren't very clear about this.

addon/components/au-time-picker.js
```diff
@@ -1,16 +1,35 @@
...
+    return formatTimeDigit([this.minuteValue]);
+  }
+  get secondValueFormatted() {
+    return formatTimeDigit([this.secondValue]);
+  }
+
+  //No-operation setters. Input changes handled by events.
+  set hourValueFormatted(newHourValue) {}
+  set minuteValueFormatted(newMinuteValue) {}
+  set secondValueFormatted(newMinuteValue) {}
```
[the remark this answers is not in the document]
benjay10: It would then complain that (e.g.) `hourValueFormatted` is a getter-only property, so to satisfy JavaScript, I created setters that don't need to do anything, but to accept some value.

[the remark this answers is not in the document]
Windvis: Hmm, I see. That's probably because of the 2-way-binding in the AuInput component then? Maybe add a todo so that we don't forget to remove these when we [swap out the Input component in the future](https://github.com/appuniversum/ember-appuniversum/issues/250)?

addon/components/au-time-picker.js
```diff
@@ -28,103 +47,44 @@ export default class AuTimePickerComponent extends Component {
...
-      if (inputValue < 0) {
-        this[elem] = 0;
-      } else if (inputValue > 59) {
-        this[elem] = 59;
-      } else {
-        this[elem] = inputValue;
-      }
-    }
+  validateTime(type, event) {
+    let tempValue = parseInt(event.target.value, 10);
```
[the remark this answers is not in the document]
benjay10: By default, some browsers do some automatic conversion, it seems. In the event that the user inputs `0x12`, the browser would default to a radix of 16 and output integer 18, and I'm not ready to deal with these weird inputs.

addon/components/au-time-picker.js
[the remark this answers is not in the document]
benjay10: Didn't think of that. That could centainly be done, but I also find that to be a little confusing. Imagine entering 63 seconds, would you then also want the minute value to go up by 1 and display the leftover 03 seconds? If you're not paying attention to the minute value that you have already entered, you might end up inserting the wrong time.

addon/components/au-time-picker.js
benjay10: Right, I saw that too, but then forgot about that! This needs to be addressed.

addon/components/au-time-picker.js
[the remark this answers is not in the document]
Windvis: This would indeed only make sense when using the arrow keys or the buttons, but it's probably not worth the complexity and requires more input from the design team. So ignore this for now, we can always tweak things in the future.

addon/components/au-time-picker.js
benjay10: This can be solved with `trackedReset`. See other comment about this. Basically, we could validate every parameter value, even on changes.

addon/components/au-time-picker.js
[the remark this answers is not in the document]
benjay10: So you mean to just add a `//TODO` with some explanation?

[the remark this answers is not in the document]
Windvis: Yea, with a link to that issue would be nice. It's strange looking code that should be removed ASAP 😄.

addon/components/au-time-picker.js
```diff
@@ -1,16 +1,52 @@
...
-  @tracked minuteValue = this.args.minutes || 0;
-  @tracked secondValue =
-    isPresent(this.args.showSeconds) && !this.args.showSeconds
-      ? 0
-      : this.args.seconds || 0; // ignore 'seconds' argument if 'showSeconds' is false
-  @tracked keyCodes = [8, 9, 13, 33, 34, 37, 39, 46];
+  @trackedReset({
+    memo() {
+      return this.validateTimeValue(this.args.hours, 'hourValue');
+    },
```
Windvis, line 11: is this custom memoize function needed? Isn't providing the path enough? It will call the validateTimeValue twice now every time an arg is changed, right?

addon/components/au-time-picker.js
```diff
@@ -1,16 +1,52 @@
...
+    update() {
+      return this.validateTimeValue(this.args.hours, 'hourValue');
+    },
+  })
+  hourValue = 12;
```
Windvis, line 16: It seems the default value is no longer used at the moment? It seems to default to 0. Does that mean that the update function also runs for the initial value?

addon/components/au-time-picker.js
```diff
@@ -1,16 +1,35 @@
...
+
+  constructor() {
+    super(...arguments);
+    this.hourValue = this.hourValue || 12;
+    this.minuteValue = this.minuteValue || 0;
+    this.secondValue = this.secondValue || 0;
+  }
+
+  get hourValueFormatted() {
+    return formatTimeDigit([this.hourValue]);
```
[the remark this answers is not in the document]
Windvis: I would still inline that util function into this file. No other component uses it and now we export another global helper for no real reason I think?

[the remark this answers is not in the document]
benjay10: Oh, I had just written integration tests for this function yesterday.

addon/components/au-time-picker.js
benjay10: I've been trying to figure this out too, but it seems like it. It works in a weird way. The documentation could be much better for this.
If the memo is a function, then the default value `= 12;` is just ignored. But when memo is a path `args.hours` then the default `= 12;` takes effect. The update function will run on initial construction, but not when the default value is taken.

addon/components/au-time-picker.js
```diff
@@ -120,4 +119,9 @@ export default class AuTimePickerComponent extends Component {
     this.secondValue = current.getSeconds();
     this.callBackParent(this.getTimeObject);
   }
+
+  formatTimeNumber(number) {
+    if (number < 10) return `0${number}`;
```
Windvis, line 124: You could also use [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) to skip the comparison.

addon/components/au-time-picker.js
```diff
@@ -1,22 +1,59 @@
...
-  @tracked minuteValue = this.args.minutes || 0;
-  @tracked secondValue =
-    isPresent(this.args.showSeconds) && !this.args.showSeconds
-      ? 0
-      : this.args.seconds || 0; // ignore 'seconds' argument if 'showSeconds' is false
-  @tracked keyCodes = [8, 9, 13, 33, 34, 37, 39, 46];
+  @trackedReset({
+    memo: 'args.hours',
+    update() {
+      if (this.args.hours) {
```
Windvis, line 10: I wonder, won't this return `12` if the parent passes in `0` as the hour value?

benjay10: Omg yes :facepalm:. I just added a test for that, and will fix this. I truly hate the truthy-falsy stuff.

addon/components/au-time-picker.js
benjay10: Nice, that would reduce this otherwise external helper to a one-liner.

addon/components/au-time-picker.js
benjay10: I propose the check `if (this.args.hours || this.args.hours == 0)` instead. This also makes "0" as a string to be a valid time value.

Windvis: I would have gone strict mode and simply not allowed something else than numbers (or undefined) and thrown assertion errors when the user did pass something invalid to the component. Strict but simpler to implement. But I don't want to hijack any more of your time so this is fine 😄.

Windvis (approved):
🎉

PR #253, opened 2022-05-11
yield the `AuToolbarGroup` component from the `AuToolbar` component

### LINKED ISSUE #251: yield `AuToolbarGroup` from the `AuToolbar` component
The `AuToolbarGroup` isn't intended to be used without a parent `AuToolbar` component so I think it would be better to yield it from the parent component instead of making it a top level component.

```hbs
<AuToolbar as |Group|>
  <Group>Some content</Group>
  <Group>Some other content</Group>
</AuToolbar>
```

We can then deprecate using the `AuToolbarGroup` component directly to guide users to the new pattern.

Windvis:
This yields the `AuToolbarGroup` component from the `AuToolbar` component and deprecates directly invoking it.

Not a breaking change, but it will create _a lot_ of noise since the component is used quite a bit 😄.

Closes #251

### DIFF
--- a/addon/components/au-accordion.hbs
+++ b/addon/components/au-accordion.hbs
@@ -4,8 +4,9 @@
     @reverse={{@reverse}}
     {{on 'click' this.openAccordion}}
     data-test-accordion-toggle
+    as |Group|
   >
-    <AuToolbarGroup>
+    <Group>
       <div>
         <AuButton
           @skin='tertiary'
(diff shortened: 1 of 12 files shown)


addon/components/au-accordion.hbs
```diff
@@ -4,8 +4,9 @@
     @reverse={{@reverse}}
     {{on 'click' this.openAccordion}}
     data-test-accordion-toggle
+    as |Group|
```
Windvis, line 7: Not sure if `ToolbarGroup` is a better name. Users can still name it like that, but most people will use the documented name I think.

Dietr (approved):

PR #259, opened 2022-05-16
Update the `AuButton` component so it matches the latest Webuniversum version

### LINKED ISSUE #254: Update the button skins
https://github.com/appuniversum/ember-appuniversum/issues/211#issuecomment-1057868350

It seems our buttons don't fully follow the Webuniversum styles and there also seems to be an extra "naked" skin type.

Dietr:
Changed the design to follow the webuniversum spec:
- added naked skin
- changed disabled and loading state

I ommited the tertiary button option as I don't see an immediate use for it. 3 button variants should offer enough hiërarchy.

### DIFF
--- a/addon/components/au-button.hbs
+++ b/addon/components/au-button.hbs
@@ -12,9 +12,17 @@
     {{/unless}}
   {{/if}}
   {{#if @hideText}}
-  <span class="au-u-hidden-visually">{{yield}}</span>
+    {{#if this.loading}}
+    <span class="au-u-hidden-visually">Aan het laden</span><AuLoader @padding="small" />
+    {{else}}
+    <span class="au-u-hidden-visually">{{yield}}</span>
+    {{/if}}
   {{else}}
-  {{yield}}
+    {{#if this.loading}}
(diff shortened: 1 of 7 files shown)


addon/components/au-button.hbs
```diff
@@ -12,9 +12,17 @@
...
+    {{/if}}
   {{else}}
-  {{yield}}
+    {{#if this.loading}}
+    Aan het laden <AuLoader @padding="small" class="au-u-margin-left-tiny" />
+    {{else}}
+    {{yield}}
+    {{/if}}
   {{/if}}
   {{#if @icon}}
```
Windvis, line 27: Should the icon still be visible if the button is in the loading state? It looks a bit strange at the moment 😄.

Windvis (commented):
Looks good. One small remark.

addon/components/au-button.hbs
Dietr: No, should be hidden 👍

Windvis (approved):

PR #261, opened 2022-05-16
Tabs component

Dietr:
Adds a tab component according to the webuniversum design:
Screenshot 2022-05-16 at 14 50 09

### DIFF
--- a/addon/components/au-tabs.hbs
+++ b/addon/components/au-tabs.hbs
@@ -1,4 +1,4 @@
-<nav class="au-c-tabs {{if @reversed 'au-c-tabs--reversed' ''}}" ...attributes>
+<nav class="au-c-tabs {{if @reversed 'au-c-tabs--reversed'}}" ...attributes>
   <ul class="au-c-tabs-list">
     {{yield}}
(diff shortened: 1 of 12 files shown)


addon/components/au-tabs.hbs
```diff
@@ -0,0 +1,5 @@
+<nav class="au-c-tabs {{if @reversed 'au-c-tabs--reversed'}}" ...attributes>
+  <ul class="au-c-tabs-list">
+    {{yield}}
```
Windvis, line 3: Impromptu review, but I would yield the component here as well, similar to the  recent AuToolbar PR. I don't think there is a use-case for using the tabs-item component directly?

Windvis:
@Dietr FYI, "enhancements" are new features so they should be released as minor versions according to the SemVer spec 😄.

addon/components/au-tabs.hbs
Dietr: Ok, will fix.

Dietr:
Sure, will follow this logic for future releases.

PR #266, opened 2022-05-19
Yield tabs item and list item

Dietr:
Fix for https://github.com/appuniversum/ember-appuniversum/pull/261#discussion_r875919373

Also fixed the list component. Not sure if I should add the deprecate here as it's a fairly new component and only used in a few places so far.


Windvis:
Thanks! Yea, if we consider it a bugfix for the previous feature I don't think we need to deprecate it indeed.

We probably do for the AuList since it was released a while ago, no?

Windvis (approved):

PR #269, opened 2022-05-30
Size option(s) for the pill component

brenner-company:
Request to add a `small` size option for the existing pill component.

### DIFF
--- a/app/styles/ember-appuniversum/_c-pill.scss
+++ b/app/styles/ember-appuniversum/_c-pill.scss
@@ -31,6 +31,10 @@ $au-pill-color-ongoing                : var(--vl-picton-120) !default;
 $au-pill-background-color-ongoing     : transparent !default;
 $au-pill-border-color-ongoing         : var(--vl-picton-120) !default;
 
+$au-pill-font-size-small              : var(--au-tiny) !default;
+$au-pill-line-height-small            : 1.3 !default;
+$au-pill-gap-small                    : .4rem !default;
+
 /* Component
   ========================================================================== */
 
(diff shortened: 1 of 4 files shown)


app/styles/ember-appuniversum/_c-pill.scss
```diff
@@ -31,6 +31,10 @@ $au-pill-color-ongoing                : var(--vl-picton-120) !default;
 $au-pill-background-color-ongoing     : transparent !default;
 $au-pill-border-color-ongoing         : var(--vl-picton-120) !default;
 
+$au-pill-font-size-small              : var(--au-tiny) !default;
```
brenner-company, line 34: @Dietr Is it ok if I leave the new size variables within this component? Or is this something that can be moved to https://github.com/appuniversum/appuniversum for example?

Dietr: Component variable are fine to put in the component. Appuniversum contains global variables and some basic setup and objects/utilitites.

Dietr (approved):

app/styles/ember-appuniversum/_c-pill.scss
brenner-company: The question was rather regarding variables such as `var(--au-tiny)` used a small(er) font-size. As you said in the previous comment, this is the correct location for it, but if I see it popping up more often in the future, is this something that can be moved to [appuniversum](https://github.com/appuniversum/appuniversum) as something like `$au-global-font-size-small`?

Dietr: We moved to root variables for font sizes and colors so I don't think we need something like `$au-global-font-size-small`. The reasoning is that for some applications these can be overridden without the need to generate sass. I probably could have made the variable name a bit more clear 😄 (eg. --au-font-size-tiny).

brenner-company: Super, thanks for the feedback!

Issue #273, opened 2022-06-02
AuModal & chrome 102


nbittich:
With the new version of chrome 102, closing the modal doesn't seem to remove the inert attribute, thus we cannot click to anything on the page (althought the modal is correctly closed).
By removing the following condition, it seems working fine:
```
    if (!this.isDestroying && this.args.modalOpen) {

```

Issue #129, opened 2021-07-27
Refactor the "double closeModal" fix


Windvis:
[focus-trap v6.6.0](https://github.com/focus-trap/focus-trap/issues/126#issuecomment-872307055) now includes the new API to [prevent the double closeModal call](https://github.com/appuniversum/ember-appuniversum/pull/111) in a better way.

Adding it as a dependency and using that new API might be a nice future refactor.

PR #274, opened 2022-06-03
Apply modal fix

Dietr:
Fix for #273 
See also https://github.com/lblod/frontend-organization-portal/commit/f3c09d80e7afc548cb28e69c8c89fa18b187ab49

@Windvis Not sure if this is undoing your previous modal fix? Can you have a look?

### DIFF
--- a/addon/components/au-modal.js
+++ b/addon/components/au-modal.js
@@ -53,18 +53,7 @@ export default class AuModal extends Component {
 
   @action
   closeModal() {
-    // Prevent that the @closeModal action is run more than once.
-    // {{focus-trap}} calls closeModal when it is deactivated which also happens when the element it is attached to is destroyed.
-    // This means that this method will be called twice if the user closes the modal with something else than the escape button.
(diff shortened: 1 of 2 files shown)


Windvis:
Yes, this reverts the fix so the `@close` action will be triggered more than once again (which is why the tests fail). I'll have to look into the problem some more I think.

addon/components/au-modal.js
```diff
@@ -38,32 +37,9 @@ export default class AuModal extends Component {
...
-    landmarkElements.forEach(function (landmarkElement) {
-      if (landmarkElement.parentElement === document.body) {
-        landmarkElement.inert = !toggle;
-      }
-    });
-  }
-
   @action
   closeModal() {
-    // Prevent that the @closeModal action is run more than once.
```
Windvis, line 56: I would keep the comment until we implement #129 cause it looks very strange without an explanation 😄.

Windvis (changes requested):

Windvis (approved):
Thanks!

PR #276, opened 2022-06-08
AuModal improvements

### LINKED ISSUE #129: Refactor the "double closeModal" fix

Windvis:
- remove the hack that was needed to prevent the double "closeModal" call (fixes #129)
- more tests

### DIFF
--- a/tests/integration/components/au-modal-test.js
+++ b/tests/integration/components/au-modal-test.js
@@ -5,6 +5,7 @@ import { hbs } from 'ember-cli-htmlbars';
 
 const MODAL = {
   ELEMENT: '[data-test-modal]',
+  BACKDROP: '[data-test-modal-backdrop]',
   HEADER: '[data-test-modal-header]',
   CLOSE: '[data-test-modal-close]',
   TITLE: '[data-test-modal-title]',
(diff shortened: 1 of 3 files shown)


Windvis:
If we want to add the inert handling in the future, I _think_ we can simply use focus-trap's onActivate and onDeactivate hooks.

Dietr (approved):

tests/integration/components/au-modal-test.js
```diff
@@ -125,14 +126,61 @@ module('Integration | Component | au-modal', function (hooks) {
...
+    };
+
+    this.isOpen = true;
+
+    await render(hbs`
+      <AuModal @modalOpen={{this.isOpen}} @closeModal={{this.handleClose}}></AuModal>
+    `);
+
+    await triggerKeyEvent(document, 'keydown', 'Escape');
+    assert.equal(timesCalled, 1, 'it closes the modal when escape is pressed');
```
Windvis, line 151: woops, that description is no longer correct.

PR #281, opened 2022-06-10
Make tabs border color consistent with toolbar, fixes #278

### LINKED ISSUE #278: AuTabs border color inconsistency
I switched over to the new `AuTabs` component from our custom implementation and I noticed that the border color doesn't match the borders of other appuniversum components. For example, if you use the AuToolbar and AuTabs next to eachother it looks like this:

image

The AuTabs version is a bit darker. I think it might make sense to make this color consistent across components (if it makes sense)?

Dietr:


app/styles/ember-appuniversum/_c-tabs.scss
```diff
@@ -9,12 +9,12 @@
   padding: 0 $au-unit;
 
   &:not(.au-c-tabs--reversed) {
-    border-bottom: .1rem solid var(--au-gray-300);
```
Windvis, line 12: Maybe this is a good opportunity to add a new css variable for this so new components can use that as well? It would make changing the border color in all components easier as well?

Dietr: Yes, we could do that. Would prefer to make a separate PR for this as this needs an appuniversum release. I would probably then look into adding some more global variables.

PR #284, opened 2022-06-10
Add data-test attributes to pill component

brenner-company:
Request to add `data-test-*` attributes to pill component.

### DIFF
--- a/addon/components/au-pill.hbs
+++ b/addon/components/au-pill.hbs
@@ -1,6 +1,10 @@
 {{#if @onClickAction}}
 <span class="au-c-pill-container">
-  <span class="au-c-pill {{this.skin}} {{this.size}} {{this.draft}}" ...attributes>
+  <span
+    class="au-c-pill {{this.skin}} {{this.size}} {{this.draft}}"
+    data-test-pill
+    ...attributes
+  >
     {{#if @icon}}
       {{#if (not-eq @iconAlignment "right")}}
       <AuIcon @icon="{{@icon}}" />
@@ -17,7 +21,12 @@
       {{/if}}
     {{/if}}
   </span>
-  <button class="au-c-pill-action {{this.actionSize}} {{this.skin}}" type="button" {{on "click" @onClickAction}}>
+  <button
+    type="button"
+    class="au-c-pill-action {{this.actionSize}} {{this.skin}}"
+    {{on "click" @onClickAction}}
+    data-test-pill-action
+  >
     <AuIcon @icon="{{@actionIcon}}" />
     <span class="au-u-hidden-visually">{{@actionText}}</span>
   </button>
@@ -29,7 +38,9 @@
     @route={{@route}}
     @models={{au-link-to-models @model @models}}
     @query={{this.queryParams}}
-    ...attributes>
+    data-test-pill
+    ...attributes
+  >
     {{#if @icon}}
       {{#if (not-eq @iconAlignment "right")}}
       <AuIcon @icon="{{@icon}}" />
@@ -50,7 +61,9 @@
   <a
     href={{@href}}
     class="au-c-pill au-c-pill--link {{this.size}} {{this.draft}}"
-    ...attributes>
+    data-test-pill
+    ...attributes
+  >
     {{#if @icon}}
       {{#if (not-eq @iconAlignment "right")}}
       <AuIcon @icon="{{@icon}}" />
@@ -68,7 +81,11 @@
     {{/if}}
   </a>
   {{else}}
-  <span class="au-c-pill {{this.skin}} {{this.size}} {{this.draft}}" ...attributes>
+  <span
+    class="au-c-pill {{this.skin}} {{this.size}} {{this.draft}}"
+    data-test-pill
+    ...attributes
+  >
     {{#if @icon}}
       {{#if (not-eq @iconAlignment "right")}}
       <AuIcon @icon="{{@icon}}" />


Dietr:
@brenner-company Can you give some context around this? Is this data-test-* something Kaleidos specific? If so, do we need to include this on all components?

brenner-company:
@Dietr No problem! We use `data-test-*` attributes (like `data-test-decision-result-pill`) for testing with Cypress. I assumed this was generally used within `ember-appuniversum` because I found several references within other components. `data-test-pill-action` is the one we really need here, because otherwise we need to target the element with either classname or element name. I've added the `data-test-pill` attributes more for clarity. If need be, I can make some adjustments 👍

Windvis:
Hmm, I wonder if this is the right way to go. Those existing data-test attributes are used by the internal testing setup. I've always considered these an implementation detail that was needed to run the local tests. I'm hesitant to officially make these "public api" since that would mean we can't change them without potentially breaking external test setups.

I think it would be better to provide new data-test attributes from the application itself. So we need to make sure all components have `...attributes` added to the outer element (or a different one that makes more sense). Another option might be to add Cypress selectors based on things that you _can_ control from the app. For the pill action button that seems to be the "action" text?

I know that isn't ideal either, but I fear allowing apps to use our internal data-test attributes will come back to bite us in the future.

Does Cypress have best practices defined for testing "third party" code which your app has no control over? In the Ember world, addons usually provide test helpers that can be used to select elements or do other things, which would hide the implementation details from consumers, maybe Cypress has something similar that we could support?

Alternatively we can introduce a policy where we consider these selectors an "semi public" api and we can only change them in minor versions and we list changes in the changelog? That way projects could define a tilde version range (`~1.1.0`) to stay on patch releases and manually bump the minor version if there are no breaking selectors?

brenner-company:
> I think it would be better to provide new data-test attributes from the application itself and use those as the base for selecting elements. We would need to make sure all components have ...attributes added to the outer element (or a different one that makes more sense).

Passing a test attribute to the component itself as a base selector (and digging deeper down yourself using element selectors like `button` or `span`) is a good idea (we are using this method now to 'fix' the issue), but wouldn't that contradict your comment in the first paragraph concerning public api and the risk of breaking external test setups? I think it would be even easier to do that?

> Another option might be to add Cypress selectors based on things that you can control from the app. For the pill action button that seems to be the "action" text?

If I understand this correctly, something like: `cy.contains('I am button text')` to select the element? It would work in theory but Cypress doesn't really recommend this (https://docs.cypress.io/guides/references/best-practices#Text-Content), the fragility regarding breaking test setups would still remain and if the, for example, `button` needs a new text, the dev would need to change it in two places (except if you would bind/name the translation string regarding to functionality rather then just subject, but again not really recommended and it would clash with the structure currently provided within Kaleidos).

> I know both options aren't ideal either, but I fear allowing apps to use our internal data-test attributes will come back to bite us in the future.

I think it is not as bad as you would initially expect? I see the chance of changing data-test attributes (that would potentially break external tests) lower then the possible options above & there is a clear separation of concern regarding naming.

> Does Cypress have best practices defined for testing "third party" code which your app has no control over?

Not that I can find at the moment. It does mention external sites (https://docs.cypress.io/faq/questions/general-questions-faq#Can-I-use-Cypress-to-script-user-actions-on-an-external-site-like-gmail-com), but I wouldn't regard the relationship Kaleidos ↔️ Ember Appuniversum as such.

The best practices section within Cypress docs has a topic on 'Selecting elements' though: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

> In the Ember world, addons usually provide test helpers that can be used to select elements or do other things, which would hide the implementation details from consumers, maybe Cypress has something similar that we could support?

Do you mean some actions like `pillAction.click()` we could expose as an API? That would be possible to call through `.invoke()` (https://docs.cypress.io/api/commands/invoke) within Cypress.

This would be a great solutions for these kind of issues, but I would have to run it by Sven (he maintains most of the tests at the moment) first. I'll keep you posted here.

In the meantime, do you have an example of this within an Ember application?

> Alternatively we can introduce a policy where we consider these selectors a "semi public" api and we can only change them in minor versions and we list changes in the changelog? That way projects could define a tilde version range (~1.1.0) to stay on patch releases and manually bump the minor version if there are no breaking selectors?

Seems like a great idea 👍 This could be a solution, but let's hold back on a decision here until we have decided the future direction from the options above.

ValenberghsSven:
This is a bit of a tough one.

>Hmm, I wonder if this is the right way to go. Those existing data-test attributes are used by the internal testing setup. I've always considered these an implementation detail that was needed to run the local tests. I'm hesitant to officially make these "public api" since that would mean we can't change them without potentially breaking external test setups.

I agree with you on this. As a dependency to Kaleidos we can use those existing selectors to test but must be wary that they are not set in stone and any update could break tests.
So adding new selectors to code you are not testing in your own internal setup does not make sense.

For some other ember dependencies, we are already using the css classes to select elements but that is also prone to breaking.
(f.e. ember-power-select)

Although it's not recommended to use css classes by cypress, we can't really expect that dependencies provide selectors for our own needs. That might lead to constant PR's for more selectors.
The recommendations are aimed towards the app that is being tested where you can add selectors wherever you want.
Other options like `id` or `name` I would just avoid unless you actively use them.

Adding `...attributes` is also not a cure-all and you might end up passing in css or html traits to wrong elements.

>In the Ember world, addons usually provide test helpers that can be used to select elements or do other things, which would hide the implementation details from consumers, maybe Cypress has something similar that we could support?
     
>Do you mean some actions like pillAction.click() we could expose as an API? That would be possible to call through .invoke() (https://docs.cypress.io/api/commands/invoke) within Cypress.

I don't know of something like that for Cypress. Don't think it fits with the general idea of Cypress. Wouldn't look further into this.
Might take way more time then you think or just not work for little to no benefits.


For right now, I would actually be inclined to not add these selectors and we will select the correct element with a combination of own test selector + dependency css.
As we gradually replace more and more local components to app-universum components (I think that's the plan?) we might gain some insight further down the line.

brenner-company:
@Windvis 

> I meant something like these helpers for ember-power-select: https://ember-power-select.com/docs/test-helpers
>
> Those are helpers that can provide you with the element and abstract the actual selector logic away. So if the underlying implementation changes, the helper can be updated as well and consuming apps don't have to know about it. So not quite the same as the Cypress's invoke. I'm not sure if that's the best solution either cause it would add substantial maintenance overhead for each component we add.

Oh, didn't know that, thanks!

-----

@ValenberghsSven 

> So adding new selectors to code you are not testing in your own internal setup does not make sense.

Wasn't thinking about it that way. Makes sense in this context 👍

> For right now, I would actually be inclined to not add these selectors and we will select the correct element with a combination of own test selector + dependency css.

I can agree with this, but just my two cents: using CSS classnames is not a good solution. 

> As we gradually replace more and more local components to app-universum components (I think that's the plan?) we might gain some insight further down the line.

That is, hopefully, the idea. Let's hope we can find a general solution on the road to that destination.

-----

Thanks for the help everyone! I'll close this PR for the meantime.

PR #302, opened 2022-08-03
ember-file-upload v5+ update

### LINKED ISSUE #222: ember-file-upload v4 deprecations
Using the `AuFileUpload` component in a v3.28 app shows the following deprecation warning:

> DEPRECATION: Using `computed.reads` has been deprecated. Instead, import the value directly from @ember/object/computed

Updating to v5 of that addon should fix it, but looking at the [upgrade guide](https://ember-file-upload.pages.dev/docs/upgrade-guide#upgrading-to-v5) that might not be trivial without triggering deprecations from the addon itself and it requires us to drop 3.24 support so it has to happen before we release v2. It would allow us to unblock Ember v4 support though.

Windvis:
This bumps ember-file-upload to v7 which required some code changes.

It's also a breaking change since it requires Ember 3.25+ so we need to release it as part of the v2 release.

~~The major change for consuming projects is that ember-file-upload [removed all validations](https://github.com/adopted-ember-addons/ember-file-upload/pull/578) and pushed that responsibility back to the consumers. At the moment I expose the `@filter` argument for this, so projects can decide if they want stricter validations, but we can also try to mimic the setup that ember-file-upload v4 used.~~ 
We reimplemented the ember-file-upload v4 validations so in theory the public API of the `AuFileUpload` component wasn't changed with this PR.

Closes #222

### DIFF
--- a/addon/components/au-file-upload.js
+++ b/addon/components/au-file-upload.js
@@ -92,9 +92,13 @@ export default class AuFileUploadComponent extends Component {
       }
     }
 
-    let hasValidFileSize = file.size < this.maxFileSizeMB * Math.pow(1024, 2);
-    if (!hasValidFileSize) {
+    if (!isValidFileSize(file.size, this.maxFileSizeMB)) {
       this.addError(file, `Bestand is te groot (max ${this.maxFileSizeMB} MB)`);
+      return false;
+    }
+
+    if (!isValidFileType(file, this.args.accept)) {
+      this.addError(file, this.helpTextFileNotSupported);
       return false;
     }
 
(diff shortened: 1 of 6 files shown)


addon/components/au-file-upload.js
```diff
@@ -67,43 +69,53 @@ export default class AuFileUploadComponent extends Component {
...
+      let shouldBeUploaded = this.args.filter(...arguments);
+
+      if (shouldBeUploaded) {
+        return true;
+      } else {
+        this.addError(file, this.helpTextFileNotSupported);
+        return false;
+      }
+    } else {
+      return true;
```
Windvis, line 111: At the moment, if no `@filter` argument is provided the file can be uploaded. This means that `@accept` will only be used by the file input to soft disable some file types in the file selector popup.

We could implement some logic to do basic file filtering based on the extension or mime type, similar to how it worked in ember-file-upload v4.

addon/components/au-file-upload.js
```diff
@@ -67,43 +69,53 @@ export default class AuFileUploadComponent extends Component {
...
+  filter(file, files, index) {
+    let isFirstFile = index === 0;
+
+    if (isFirstFile) {
+      this.resetErrors();
+    } else {
+      if (!this.args.multiple) {
+        // We only upload the first file if `@multiple` is not set to true. This matches the behavior of ember-file-upload v4.
+        return false;
+      }
```
Windvis, line 92: I mimicked ember-file-upload v4's behavior where only the first file is uploaded if the users drops more than 1 and multiple is set to false. I'm not sure if this is the best behavior though. It might be better to not upload anything at all in that case?

addon/components/au-file-upload.js
Windvis: As discussed in Rocket.chat, making this as frictionless as possible for projects (and end-users) is preferred, so we should try to add a basic mime-type / extension based validation system similar to how it worked in ember-file-upload v4.

addon/components/au-file-upload.js
```diff
@@ -67,43 +69,57 @@ export default class AuFileUploadComponent extends Component {
...
+
+      if (shouldBeUploaded) {
+        return true;
+      } else {
+        this.addError(file, this.helpTextFileNotSupported);
+        return false;
+      }
+    } else {
+      return true;
+    }
```
Windvis, line 116: At the moment it is possible to add extra validations on top by passing the `@filter` argument. The current implementation doesn't let you return a validation message though. It sounds nice to allow that? 

Some options:
- let `@filter` return a boolean _or_ a string. `true` = valid, `false` = invalid (use current logic), `"string"` = invalid, use custom message
- let `@filter` return an object `{ isValid: boolean, error?: string }`

Any preferences or other suggestions?

Windvis: It might also be better to rename `@filter` to something else since it doesn't fully match the `ember-file-upload` version anymore.

Windvis: Or we drop the `@filter` functionality for now since it's a now feature and this was just meant to be an update. I initially didn't implement the v4 validations so the filter was the alternative, but now that's not needed anymore. I'll go that route for now and open a feature request issue for the custom validations.

PR #313, opened 2022-08-24
Various small improvements to button component

brenner-company:
The following changes should allow easier integration of button component within the Kaleidos (or any) app:
- Switch regarding spacing (inside button) from margins to gap: enables automatic spacing of children + fixes icon-only redundant spacing issue.
- Add `au-c-button--icon-only` class if an icon is defined and @hideText is set to true: allows for additional styling if needed (not applied by default). As used within the link component. E.g. inside the Kaleidos app, we need less horizontal padding on icon-only buttons.

Also added testing to cover the changes mentioned above.

### DIFF
--- a/app/styles/ember-appuniversum/_s-root.scss
+++ b/app/styles/ember-appuniversum/_s-root.scss
@@ -125,6 +125,11 @@
   --au-h1-medium: 4rem;
   --au-h1: 4.4rem;
 
+  // Icons
+  --au-icon-size: 1.3rem;
+  --au-icon-size-medium: 1.6rem;
+  --au-icon-size-large: 1.9rem;
+
   // UI
   --au-radius: .3rem;
   --au-border: .2rem;
--- a/tests/integration/components/au-button-test.js
+++ b/tests/integration/components/au-button-test.js
@@ -24,18 +24,6 @@ module('Integration | Component | au-button', function (hooks) {
     assert.dom(this.element).hasText('template block text');
   });
 
-  test('it hides the text when @hideText is set to true', async function (assert) {
-    await render(hbs`
-      <AuButton
-        @hideText={{true}}
-      >
-        template block text
-      </AuButton>
-    `);
-
-    assert.dom('.au-u-hidden-visually').exists();
-  });
-
   test('it adds the icon-only modifier class when an icon is defined and @hideText is set to true', async function (assert) {
     await render(hbs`
       <AuButton
--- a/tests/integration/components/au-button-test.js
+++ b/tests/integration/components/au-button-test.js
@@ -23,17 +23,4 @@ module('Integration | Component | au-button', function (hooks) {
 
     assert.dom(this.element).hasText('template block text');
   });
-
-  test('it adds the icon-only modifier class when an icon is defined and @hideText is set to true', async function (assert) {
-    await render(hbs`
-      <AuButton
-        @icon="plus"
-        @hideText={{true}}
-      >
-        template block text
-      </AuButton>
-    `);
-
-    assert.dom('button').hasClass('au-c-button--icon-only');
-  });
 });
--- a/tests/integration/components/au-button-test.js
+++ b/tests/integration/components/au-button-test.js
@@ -23,4 +23,29 @@ module('Integration | Component | au-button', function (hooks) {
 
     assert.dom(this.element).hasText('template block text');
   });
+
+  test('it hides the text when @hideText is set to true', async function (assert) {
+    await render(hbs`
+      <AuButton
+        @hideText={{true}}
+      >
+        template block text
+      </AuButton>
+    `);
+
+    assert.dom('.au-u-hidden-visually').exists();
+  });
+
+  test('it adds the icon-only modifier class when an icon is defined and @hideText is set to true', async function (assert) {
+    await render(hbs`
+      <AuButton
+        @icon="plus"
+        @hideText={{true}}
+      >
+        template block text
+      </AuButton>
+    `);
+
+    assert.dom('button').hasClass('au-c-button--icon-only');
+  });
 });
--- a/addon/components/au-button.hbs
+++ b/addon/components/au-button.hbs
@@ -4,33 +4,26 @@
   ...attributes
 >
   {{#unless this.loading}}
-    {{#if @icon}}
-      {{#if (eq @iconAlignment "left")}}
-      <AuIcon @icon="{{@icon}}" @alignment="left" />
-      {{/if}}
-      {{#unless @iconAlignment}}
+    {{#if (and @icon (eq this.iconAlignment "left"))}}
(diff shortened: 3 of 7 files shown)


tests/integration/components/au-button-test.js
```diff
@@ -23,4 +23,29 @@ module('Integration | Component | au-button', function (hooks) {
...
+    await render(hbs`
+      <AuButton
+        @icon="plus"
+        @hideText={{true}}
+      >
+        template block text
+      </AuButton>
+    `);
+
+    assert.dom('button').hasClass('au-c-button--icon-only');
```
brenner-company, line 49: Is using `.dom('button')` good practice by the way or should I use a data-test selector here?

Windvis: I think it's fine for simple tests like this and in places where the element isn't an implementation detail of the component 👍.

tests/integration/components/au-button-test.js
```diff
@@ -23,4 +23,29 @@ module('Integration | Component | au-button', function (hooks) {
 
     assert.dom(this.element).hasText('template block text');
   });
+
+  test('it hides the text when @hideText is set to true', async function (assert) {
```
Windvis, line 27: Even though I appreciate it _a lot_ that you added tests 🎉 , I think these kind of tests aren't super useful in general 😅. Checking for certain classes isn't a good practice (imo) since it's an implementation detail and it doesn't guarantee that the applied css actually makes the text hidden.

We actually want to test if things are visually hidden, which we now can because of the Chromatic setup that Dieter added a while ago (#267).

I'm not 100% sure, but I think Chromatic makes screenshots based on the stories. So I think we can add a new story for the "Icon only" button and that way we will know for certain if something breaks the text hiding in the future.

Windvis (commented):
(some quick feedback about the tests)

Dietr (approved):

tests/integration/components/au-button-test.js
brenner-company: > Even though I appreciate it _a lot_ that you added tests 🎉 , I think these kind of tests aren't super useful in general 😅. Checking for certain classes isn't a good practice (imo) since it's an implementation detail and it doesn't guarantee that the applied css actually makes the text hidden.

Wasn't absolutely sure about that one too 🫣 I kept looking for an actual solution to check for visibility as `isVisible` only checks for the element dimensions, but couldn't find anything.

> We actually want to test if things are visually hidden, which we now can because of the Chromatic setup that Dieter added a while ago (#267).
> 
> I'm not 100% sure, but I think Chromatic makes screenshots based on the stories. So I think we can add a new story for the "Icon only" button and that way we will know for certain if something breaks the text hiding in the future.

This seems like a way better solution 👍 I'll give it a look when I have the time. Thanks for mentioning it!

@Windvis I'll delete this test, but is the last one I added worth keeping?

Windvis: Hmm, I'm not sure if I'm a fan of that class to be honest 😅. Making css classes part of the components "public API" of a component is not something I like personally, but I've seen it done before.

I would probably wrap the component in the project and add the class myself. Something like `<IconButton @icon="something">Hidden text</IconButton>` which would set the `@hideText` arg and the icon-only class you need for your projects styling, and passes through all the other arguments the button supports.

If it is actually considered public API then yes I think the test makes sense.

Windvis: > This seems like a way better solution 👍 I'll give it a look when I have the time. Thanks for mentioning it!

I _think_ it is as simple as adding a new [named export ](https://github.com/appuniversum/ember-appuniversum/blob/f8118bcfe5880e94a0aae9a0d2c62b361532cefa/stories/5-components/Buttons/AuButton.stories.js#L117-L129) with the proper arg values.

brenner-company: > Hmm, I'm not sure if I'm a fan of that class to be honest 😅. Making css classes part of the components "public API" of a component is not something I like personally, but I've seen it done before.
> 
> I would probably wrap the component in the project and add the class myself. Something like `<IconButton @icon="something">Hidden text</IconButton>` which would set the `@hideText` arg and the icon-only class you need for your projects styling, and passes through all the other arguments the button supports.
> 
> If it is actually considered public API then yes I think the test makes sense.

I think I'll hand over the decision on that to @Dietr. Do you have a taken on this?

brenner-company: As an other option, I could also make the visual changes for the icon-only button (so it has the same width as height) here inside `ember-appuniversum`, but I'm not sure this is a common request. As a plus, we could also size up the icon a bit as it is done inside `AuLink`. https://github.com/appuniversum/ember-appuniversum/blob/b73e76e8ff51713065c703e05194b2bfa08ee8a2/app/styles/ember-appuniversum/_c-link.scss#L81

brenner-company: I did a quick check in our project and the icon inside the icon-only version of the button is really small at the moment.

(on the left: our own component - on the right: `AuButton`)
CleanShot 2022-08-26 at 10 58 06@2x

Maybe I could kill two birds with one stone and:
- Adjust the icon size inside the icon-only button
- Adjust horizontal padding of the icon-only button here inside `ember-appuniversum` or use the `au-c-button--icon-only` modifier in our own project.

Thoughts?

Windvis: I found this comment by Dieter: https://github.com/lblod/ember-submission-form-fields/blob/46af3ccef7a0045546a02cdc0c4ca311fe30ba77/addon/components/rdf-input-fields/concept-scheme-multi-select-checkboxes.hbs#L20-L21

It seems Webuniversum does have an icon only variant, so if the Appuniversum version doesn't look like that I assume it makes sense to fix it here?

brenner-company: So to check off ✅: the styling changes are ok, but remove the remaining test I added in favour of a visual regression test?

Windvis: That's what I would do, yea, but @Dietr might have a different vision?

Dietr: I'm fine with the changes you propose. A bigger icon in the icon only button seems like a good improvement for all projects 👍

app/styles/ember-appuniversum/_s-root.scss
```diff
@@ -125,6 +125,11 @@
   --au-h1-medium: 4rem;
   --au-h1: 4.4rem;
 
+  // Icons
```
brenner-company, line 128: @Dietr Is it ok to refactor this as CSS variables? Not entirely sure about this one as SCSS variables should also work with swapping the import order of components a bit.

 You can check `_c-icon.scss` and `_c-button.scss` for the implementation.

brenner-company:
@Dietr Bedankt voor de hulp! Zou een release ergens één van de volgende dagen mogelijk zijn?

Issue #318, opened 2022-09-08
PowerSelectMultiple disabled state padding issue


Windvis:
**Describe the bug**
It seems the selected options are missing some whitespace when the component is in its "disabled" state.

![image](https://user-images.githubusercontent.com/3533236/189126105-95663538-0791-4f5e-8cec-45899df99417.png)

I took a quick look, and it seems that the removal button contained the needed whitespace but that button is hidden in the disabled state.

**To Reproduce**
1. Use a PowerSelectMultiple component
2. Add some selected options
3. set the `@disabled` argument to `true

**Expected behavior**
The selected items have proper whitespace

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOs
 - Browser: FireFox
 - Version: 105

Windvis:
Closed by #320

PR #320, opened 2022-09-12
Fix power select multiple disabled state, #318

Dietr:
Fixed the padding on the pill when the power select is disabled. 
Also added a story for the multiple power select but having some trouble to demo the disabled state properly.


Windvis:
I poked at the stories and I think we need to create a new file for the PowerSelectMultiple example since the 'selected' type is an array instead of a string.

I created #321 for that (which targets this branch)

Windvis (approved):

PR #322, opened 2022-09-13
Remove deprecated code

Windvis:
This removes all deprecation warnings and the code that relates to it.


Windvis:
Whoops, it seems we still have some AuToolbarGroup usage in the stories 😬. Good catch by Chromatic!

PR #323, opened 2022-09-16
Rework dropdown

Dietr:
- use AuButton component
- allow <AuHr /> component as a divider element
- simplify component

### DIFF
--- a/addon/components/au-dropdown.hbs
+++ b/addon/components/au-dropdown.hbs
@@ -1,4 +1,5 @@
 <div class="au-c-dropdown" ...attributes>
+  {{#if this.dropdownOpen}}
   <AuButton
     @skin={{this.skin}}
     @size={{@size}}
@@ -8,12 +9,28 @@
     @alert={{@alert}}
     aria-haspopup="true"
     aria-expanded="{{if this.dropdownOpen "true" "false"}}" type="button"
-    {{on "click" this.toggleDropdown}}>
+    {{on "click" this.closeDropdown}} data-test-dropdown-button>
     <span class="au-c-dropdown__name" data-test-dropdown-title>
       {{this.title}}
     </span>
   </AuButton>
-  <div id="{{this.id}}" class="au-c-dropdown__menu {{this.alignment}} {{if this.dropdownOpen "is-visible" ""}}" role="menu" tabindex="-1" {{focus-trap isActive=this.dropdownOpen shouldSelfFocus=true focusTrapOptions=(hash onDeactivate=this.toggleDropdown clickOutsideDeactivates=true)}} {{on "click" this.toggleDropdown}}>
+  {{else}}
+  <AuButton
+    @skin={{this.skin}}
+    @size={{@size}}
+    @icon={{this.icon}}
+    @iconAlignment={{this.iconAlignment}}
+    @hideText={{@hideText}}
+    @alert={{@alert}}
+    aria-haspopup="true"
+    aria-expanded="{{if this.dropdownOpen "true" "false"}}" type="button"
+    {{on "click" this.openDropdown}} data-test-dropdown-button>
+    <span class="au-c-dropdown__name" data-test-dropdown-title>
+      {{this.title}}
+    </span>
+  </AuButton>
+  {{/if}}
+  <div id="{{this.id}}" class="au-c-dropdown__menu {{this.alignment}} {{if this.dropdownOpen "is-visible" ""}}" role="menu" tabindex="-1" {{focus-trap isActive=this.dropdownOpen shouldSelfFocus=true focusTrapOptions=(hash onDeactivate=this.closeDropdown clickOutsideDeactivates=true)}} {{on "click" this.closeDropdown}}>
     {{yield}}
   </div>
 </div>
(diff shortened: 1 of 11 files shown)


addon/components/au-dropdown.js
```diff
@@ -60,6 +61,15 @@ export default class AuDropdown extends Component {
 
       return this.args.dropdownButtonLabel;
     } else if (this.args.buttonLabel) {
+      deprecate('@dropdownLabel is deprecated', false, {
+        id: '@appuniversum/ember-appuniversum.au-dropdown.dropdownButtonLabel-argument',
+        until: '2.0.0',
+        for: '@appuniversum/ember-appuniversum',
+        since: {
+          enabled: '1.6.0',
```
Windvis, line 69: Should be 1.7.0 I think since 1.6.1 has already been released.

brenner-company:
Looks good 👍

As a finishing touch, I made a tiny adjustment to the focus styling of the menu items so that only the first & last item get rounded corners.

Just out of interest: does the `role="group"` have a specific function within the dropdown component?

brenner-company (approved):

addon/components/au-dropdown.js
```diff
@@ -60,6 +61,15 @@ export default class AuDropdown extends Component {
 
       return this.args.dropdownButtonLabel;
     } else if (this.args.buttonLabel) {
+      deprecate('@dropdownLabel is deprecated', false, {
+        id: '@appuniversum/ember-appuniversum.au-dropdown.dropdownButtonLabel-argument',
+        until: '2.0.0',
```
Windvis, line 66: Do we still want to target v2 since that's almost ready to go? Maybe we should target v3 instead so consumers have more time to remove the `dropdownLabel` usage?

I'm not sure how widely used it is so maybe it's not needed to keep it around for another major version.

addon/components/au-dropdown.hbs
```diff
@@ -1,25 +1,36 @@
...
+  <AuButton
+    @skin={{this.skin}}
+    @size={{@size}}
+    @icon={{this.icon}}
+    @iconAlignment={{this.iconAlignment}}
+    @hideText={{@hideText}}
+    @alert={{@alert}}
+    aria-haspopup="true"
+    aria-expanded="{{if this.dropdownOpen "true" "false"}}" type="button"
+    {{on "click" this.closeDropdown}} data-test-dropdown-button>
```
Windvis, line 12: The only difference between the if and the else version seems to be the action, so I think it can be replaced with something like this?

```suggestion
    {{on "click" (if this.dropdownOpen this.closeDropdown this.openDropdown)}} data-test-dropdown-button>
```

Or alternatively with something like this if we add a toggle action:

```suggestion
    {{on "click" this.toggleDropdown}} data-test-dropdown-button>
```

addon/components/au-dropdown.hbs
```diff
@@ -1,25 +1,36 @@
...
+    @alert={{@alert}}
+    aria-haspopup="true"
+    aria-expanded="{{if this.dropdownOpen "true" "false"}}" type="button"
+    {{on "click" this.openDropdown}} data-test-dropdown-button>
+    <span class="au-c-dropdown__name" data-test-dropdown-title>
+      {{this.title}}
+    </span>
+  </AuButton>
+  {{/if}}
+  <div id="{{this.id}}" class="au-c-dropdown__menu {{this.alignment}} {{if this.dropdownOpen "is-visible" ""}}" role="menu" tabindex="-1" {{focus-trap isActive=this.dropdownOpen shouldSelfFocus=true focusTrapOptions=(hash onDeactivate=this.closeDropdown clickOutsideDeactivates=true)}} {{on "click" this.closeDropdown}}>
```
Windvis, line 33: Is the `id` attribute needed? It seems like nothing links to it?

addon/components/au-dropdown.js
```diff
@@ -14,13 +14,14 @@ export default class AuDropdown extends Component {
   // Open dropdown
   @action
   openDropdown() {
-    // Toggle dropdown view state
+    // Open dropdown view state
```
Windvis, line 17: Nitpick: I would simply delete these kind of comments since they don't add much value over the method name anyways 😄.

addon/components/au-dropdown.js
Windvis: Also, it's not very clear what action the user should take to resolve the deprecation. Should they simply remove the argument?

addon/components/au-dropdown.hbs
Dietr: I tried to do this but creates a conflict with the focus-trap deactivate when you want to close the dropdown by clicking the button. Not sure how to solve it.

Dietr:
> Looks good 👍
> 
> As a finishing touch, I made a tiny adjustment to the focus styling of the menu items so that only the first & last item get rounded corners (see screenshot below).
> 
> CleanShot 2022-09-19 at 15 38 47@2x
> 
> Just out of interest: does the `role="group"` have a specific function within the dropdown component?

It can be used to group menu items. It's a valid role within the menu.

addon/components/au-dropdown.hbs
Windvis: I took a look at it and it seems we were simply sidestepping the issue by forcing a rerender with a different button component which meant the button click handler was ignored due to the teardown.

I created this [PR](https://github.com/appuniversum/ember-appuniversum/pull/324) which solves it in a different way and allows us to remove the duplication.

PR #324, opened 2022-09-21
AuDropdown refactoring

Windvis:
This removes the duplicated AuButton. We now use focus-trap's clickOutsideDeactivates event to detect when an outside click happened, and only call the close method if the target element is not the toggle button. This prevents the dropdown from instantly reopening again. The previous implementation worked around this by rerendering a different button component which meant the click event was ignored.

### DIFF
--- a/addon/components/au-dropdown.js
+++ b/addon/components/au-dropdown.js
@@ -16,6 +16,23 @@ export default class AuDropdown extends Component {
     this.dropdownOpen = false;
   }
 
+  @action
+  toggleDropdown() {
+    this.dropdownOpen = !this.dropdownOpen;
+  }
+
+  @action
+  clickOutsideDeactivates(event) {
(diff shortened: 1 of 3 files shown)


addon/components/au-dropdown.js
```diff
@@ -16,6 +16,23 @@ export default class AuDropdown extends Component {
...
   }
 
+  @action
+  toggleDropdown() {
+    this.dropdownOpen = !this.dropdownOpen;
+  }
+
+  @action
+  clickOutsideDeactivates(event) {
+    let toggleButton = document.querySelector('[data-au-dropdown-toggle]');
```
Windvis, line 26: We can replace this with a local modifier so we don't need to add the data attribute, but that requires Ember 3.25+ so we can only do that in the v2 release.

Dietr (approved):

PR #328, opened 2022-10-05
Faulty vertical alignment of button & link components

brenner-company:
The following issue is visible when two or more successive `AuButton` or `AuLink` components with one having an icon with left alignment and the other having an icon with left alignment (or even no icon) are displayed: the elements will be vertically aligned in an incorrect way (see screenshot below).

CleanShot 2022-09-14 at 14 31 43@2x


brenner-company:
@Dietr I'm putting this PR on hold for a moment. I'm seeing some minor visual (vertical) jump issues within Chromatic that I didn't experience within Kaleidos.

Issue #339, opened 2022-11-09
Additional skins for external link component


brenner-company:
Within Kaleidos we need to a few links for downloading files that are styled as naked buttons (for example, as used for the [regular link component](https://appuniversum.github.io/ember-appuniversum/?path=/story/components-links-aulink--naked-button&args=icon:download;hideText:true)). Because the regular link component is a `LinkTo` it doesn't accept `download` attributes and we can't use it in this case.

Can I add the 3 (missing) `button-` skins (`button`, `button-secondary` & `button-naked`) to the `AuLinkExternal` component?

This FR is rather urgent because it is a bug release for Kaleidos will be waiting on this.

Windvis:
We could also use this for Digitaal Loket. I noticed we've been using regular `a` tags with the needed classes in some places so this would be a good replacement for that 👍.

Windvis:
Closed by #340

PR #319, opened 2022-09-09
Spacing issue (icon + text) link component with button skin

brenner-company:
The changes within https://github.com/appuniversum/ember-appuniversum/pull/313 caused an unintended issue: when the AuLink component is displayed with either a `button`, `button-secondary` or `button-naked` skin there is additional whitespace between the icon and the text.

image

Preview URL: https://appuniversum.github.io/ember-appuniversum/?path=/story/components-links-aulink--component&args=skin:button

This issue was not directly caught because the AuLink component with a button-like skin is currently not covered by the Chromatic tests.

The changes I made:
- Performed the same structural (& functional) changes on AuLink and AuLinkExternal made within the AuButton component (see: https://github.com/appuniversum/ember-appuniversum/pull/313/files#diff-08dd95d3f03a9928ab26dce9607419e9c22e141e50a28e916ba9592ac5ca1713).
- Enabled gap spacing within AuLink styling.
- Added stories for AuLink similar to AuButton (so it is covered by Chromatic tests).

Issue #219, opened 2022-03-24
Release new versions from CI


Windvis:
It would be nice if we could[ configure Github Actions](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry) to do the npm publishing, that way people with the right Github permissions can simply push the tag and not everyone needs npm permissions. This would unblock us if an urgent fix needs to be released but people are unavailable.

To clarify, a release would work like this:

1. we run release-it locally
2. release it will generate a changelog (lerna-changelog)
3. release-it will bump the version number and create a new tag
4. the tag is pushed to GitHub and a new GitHub release is created
5. GitHub Actions picks up the new tag and publishes a new version to the npm registry

We use a similar setup for some of the lblod addons (but it uses Drone for CI).

PR #340, opened 2022-11-09
Addition of button skins for external link component

brenner-company:
The addition of 3 (missing) `button-` skins (`button`, `button-secondary` & `button-naked`) to the `AuLinkExternal` component.

See https://github.com/appuniversum/ember-appuniversum/issues/339 for additional information.

### DIFF
--- a/addon/components/au-link-external.js
+++ b/addon/components/au-link-external.js
@@ -3,10 +3,13 @@ import Component from '@glimmer/component';
 const SKIN_CLASSES = {
   primary: 'au-c-link',
   secondary: 'au-c-link au-c-link--secondary',
+  button: 'au-c-button au-c-button--primary',
+  'button-secondary': 'au-c-button au-c-button--secondary',
+  'button-naked': 'au-c-button au-c-button--naked',
 };
 
 export default class AuLinkExternal extends Component {
-  get skin() {
+  get skinClass() {
     if (SKIN_CLASSES[this.args.skin]) {
       return SKIN_CLASSES[this.args.skin];
     } else {
@@ -14,13 +17,28 @@ export default class AuLinkExternal extends Component {
     }
   }
 
-  get width() {
-    if (this.args.width == 'block') return 'au-c-link--block';
+  get widthClass() {
+    if (this.args.width == 'block')
+      if (this.args.skin.startsWith('button')) return 'au-c-button--block';
+      else return 'au-c-link--block';
+    else return '';
+  }
+
+  get activeClass() {
+    if (this.args.active) return 'is-active';
     else return '';
   }
 
   get iconAlignment() {
     if (this.args.iconAlignment) return this.args.iconAlignment;
     else return 'left';
+  }
+
+  get iconOnlyClass() {
+    if (this.args.icon && this.args.hideText)
+      if (this.args.skin && this.args.skin.startsWith('button'))
+        return 'au-c-button--icon-only';
+      else return 'au-c-link--icon-only';
+    return '';
   }
 }
--- a/addon/components/au-link-external.js
+++ b/addon/components/au-link-external.js
@@ -24,11 +24,6 @@ export default class AuLinkExternal extends Component {
     else return '';
   }
 
-  get activeClass() {
-    if (this.args.active) return 'is-active';
-    else return '';
-  }
-
   get iconAlignment() {
     if (this.args.iconAlignment) return this.args.iconAlignment;
     else return 'left';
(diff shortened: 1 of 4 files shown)


addon/components/au-link-external.js
```diff
@@ -3,24 +3,42 @@ import Component from '@glimmer/component';
...
-  get width() {
-    if (this.args.width == 'block') return 'au-c-link--block';
+  get widthClass() {
+    if (this.args.width == 'block')
+      if (this.args.skin.startsWith('button')) return 'au-c-button--block';
+      else return 'au-c-link--block';
+    else return '';
+  }
+
+  get activeClass() {
```
Windvis, line 27: I don't think we need an "active" state for an external link? At least not one based on an argument.

stories/5-components/Links/AuLinkExternal.stories.js
```diff
@@ -61,3 +67,64 @@ Component.args = {
...
+  route: 'index',
+  skin: 'secondary',
+  icon: 'manual',
+  iconAlignment: 'left',
+  hideText: false,
+  width: '',
+  active: false,
+};
+Secondary.parameters = {
+  chromatic: { disableSnapshot: true },
```
Windvis, line 84: Is there a specific reason why we don't want Chromatic to snapshot these variants?

Windvis (commented):
Some questions but other than that it looks good to me 👍.

stories/5-components/Links/AuLinkExternal.stories.js
brenner-company: This was mostly because of a previous comment (in another PR): https://github.com/appuniversum/ember-appuniversum/pull/319#discussion_r967063270

If you think this wont impact the runtime (of the tests) too much, we can opt to enable them again?

addon/components/au-link-external.js
brenner-company: In the previous state of the component this did nothing because no string was bound to `this.active` so I assumed that this was a bug? Plus the fact that the active state was also toggle-able (but with no visual effect) inside Storybook. Maybe @Dietr can enlighten us?

Windvis: I assumed it was a copy pasta mistake from the AuLink component actually 😄.

stories/5-components/Links/AuLinkExternal.stories.js
Windvis: I would honestly enable it regardless if it impacts the test time. It seems like a good thing to test visually since we can't really verify if it works through other tests. And I would be surprised if it had _that_ much impact?

addon/components/au-link-external.js
brenner-company: Well, that also makes sense 😅 I'll remove this.

stories/5-components/Links/AuLinkExternal.stories.js
brenner-company: The impact was more regarding the future if the grand total of components (and thus tests) gets larger. More like a mindset to not test everything in absolute detail.

But I also agree to your point of view. Maybe we'll enable the tests here (and for `AuLink`) again, but keep an eye out if the runtime gets too long (if that is even possible 😅)?

Windvis: Yea, I'm not super familiar with Chromatic but I can't find how long it took to verify all the snapshots. Only the build time but that's only 15 seconds.

In any case, I think will always be the slowest part of our CI so I don't think we need to be conservative with Chromatic for now. We might need to look into it if we start into the free plan limitations but I don't think that will happen soon.

app/styles/ember-appuniversum/_c-link.scss
```diff
@@ -43,13 +41,6 @@ $au-link-secondary-active-color          : var(--au-gray-900) !default;
     outline: var(--au-outline);
     outline-offset: var(--au-outline-offset);
   }
-
-  &.is-active {
```
Windvis, line 47: I don't think these styles should be removed? It affects the `AuLink` component as well and there it does make sense.

Windvis (changes requested):

app/styles/ember-appuniversum/_c-link.scss
brenner-company: Oh! I was a little fast there. I'll revert those changes!

Windvis (approved):

Windvis:
LGTM! I can't create a release myself though (which reminds me that I need to finish #219 😅).

Windvis:
Released as part of [v1.10.0](https://github.com/appuniversum/ember-appuniversum/releases/tag/v1.10.0)!

PR #341, opened 2022-11-10
Use Github Actions to publish to npm

### LINKED ISSUE #219: Release new versions from CI

Windvis:
Publishing from CI has the advantage that everyone with write permissions to the repo can release a new version without having to have npm rights to the package.

Closes #219

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,22 @@
+name: 'Publish to npm'
+on:
+  push:
+    tags:
+       - "v*.*.*"
+
+jobs:
+  publish:
+    name: 'Publish'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 12.x
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
--- a/package.json
+++ b/package.json
@@ -155,6 +155,9 @@
     "github": {
       "release": true,
       "tokenRef": "GITHUB_AUTH"
+    },
+    "npm": {
+      "publish": false
     }
   },
   "volta": {


.github/workflows/publish.yml
```diff
@@ -0,0 +1,22 @@
+name: 'Publish to npm'
+on:
+  push:
+    tags:
+       - "v*.*.*"
```
Windvis, line 5: Docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

Dietr:
> `NODE_AUTH_TOKEN`

Added the NODE_AUTH_TOKEN to the secrets.

Windvis:
Cool, thanks. I haven't tried if this actually works since it's hard to test. Maybe I'll try it out in one of my own repos first before rolling it out here.

Windvis:
I've tried this setup in the [ember-router-service-refresh-polyfill addon](https://github.com/ember-polyfills/ember-router-service-refresh-polyfill) and it seems to [work as expected](https://github.com/ember-polyfills/ember-router-service-refresh-polyfill/actions/runs/3472628271/jobs/5803700634). So I think we're good 👍.

Windvis:
For future readers: We also needed to [install the dependencies before publishing](https://github.com/appuniversum/ember-appuniversum/commit/bc6423bda47259f689d245ca3d5b27c472dbdef4) since we're using `prepare` scripts that expect the dependencies to be installed.

PR #343, opened 2022-11-15
Provide explicit unit to `mix` builtin

abeforgit:
as per https://sass-lang.com/documentation/breaking-changes/function-units#weight the color.mix function now needs an explicit unit for the third argument.
This commit applies the suggested fix

### DIFF
--- a/app/styles/ember-appuniversum/_s-colors.scss
+++ b/app/styles/ember-appuniversum/_s-colors.scss
@@ -6,7 +6,7 @@
 // ---
 // Makes a color lighter
 @function tint($color, $percent) {
-  @return mix(#fff, $color, $percent);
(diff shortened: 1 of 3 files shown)


Dietr (approved):

Windvis:
It seems the Chromatic check is failing because the PR is made from a fork and Github limits access to the repository secrets because of that:

> Secrets are environment variables that are encrypted. Anyone with collaborator access to this repository can use these secrets for Actions.
>
> Secrets are not passed to workflows that are triggered by a pull request from a fork. [Learn more about encrypted secrets](https://docs.github.com/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets).

Ideally we also run the visual regression tests for "third party" contributors but I'm not sure how to accomplish that 🤔.

Issue #333, opened 2022-10-19
More dynamic display of menu element within dropdown component 


brenner-company:
**After replacing our own dropdown component with `AuDropdown` we encountered the following (non-breaking) issue:**
When using a dropdown within a container with a defined overflow property (`auto`, `scroll`, etc.) we noticed that the content of that container jumped position to display the menu when there is insufficient space. Also, when the trigger for that dropdown is for example at the bottom of a container (or the page in general) with its overflow hidden,  the menu can be clipped even if there is enough space on top of the trigger.

A example of the first issue can be found below:

![CleanShot 2022-10-19 at 11 57 26@2x](https://user-images.githubusercontent.com/18174827/196714060-62b86ff1-4fe7-4edc-9fc5-190d61db0e3a.png)

![CleanShot 2022-10-19 at 11 57 40@2x](https://user-images.githubusercontent.com/18174827/196720490-0819860d-9e5d-4c4f-8b9d-cc81d5ecb478.png)

**I think the cause of the issues here is:**
-  the fact that the menu is displayed alongside the trigger (inside the HTML) instead of at the end of the `body` element for example.
- the fact that the positioning the menu is done solely by preset and doesn't have functionality to calculate a back-up position if there is insufficient space.

These are two things what most external popover/tooltip plugins seem to provide.

Is this something that we could provide ourselves? Or could we use something like https://github.com/kybishop/ember-attacher (we used this within Kaleidos for our own dropdown component) to enable this?

brenner-company:
I made a quick prove of concept (of dynamic positioning) using Popper JS (https://popper.js.org/ with https://github.com/NullVoxPopuli/ember-popperjs) and it seems to work well. Could this be an option?

See following file changes: https://github.com/appuniversum/ember-appuniversum/compare/feature/dynamic-display-of-menu-within-dropdown

Don't mind the bad JS and commented SCSS for the moment 😉

brenner-company:
There also seems to be an option to attach the menu to the `body` element (this fixes the container jump issue when `auto`, `scroll`, etc. is defined), but I don't have any time left today to give it a look yet.

Windvis:
I think using something like popper could be a good idea, yes 👍. It could probably be useful for other features in the future as well. The only problem at the moment is that `ember-popperjs` requires Ember v3.25+ and we still support v3.24 in Appuniversum v1. So we can't release it as part of v1 like that.

(I should just finish up the v2 work which bumps the support to v3.28).

[...]

PR #344, opened 2022-11-18
Addition of floating UI helpers

brenner-company:
The addition of a floating UI modifier & component for further use within other components that use popovers, dropdowns, etc.

The modifier & component are created with the use of https://floating-ui.com/.

**This component (& modifier in a more limited way) can:**
- automatically position a floater according to a placement value
- flip the floater if there is insufficient space on the default placement side
- negate clipping by a (parent) container with a defined overflow property (`auto`, `scroll`, etc.) by utilising a `fixed` position
- automatically position and flip a predefined arrow element according to the placement value
- automatically hide the floater when the reference element is out of view/bounds

**What still must be done is:**
- [x] create a Storybook add-on that provides the scrolling/overflow preview I included within the story (reference element needs to be centered within container when the story is loaded)
- [x] move example styling of reference, floater & arrow to a better location?
- [x] provide correct tests for the modifier & component

See #333 for additional information.


addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,158 @@
...
+import {
+  autoUpdate,
+  computePosition,
+  flip,
+  hide,
+  offset,
+  arrow,
+} from '@floating-ui/dom';
+
+import { merge } from 'merge-anything';
```
brenner-company, line 14: I needed some functionality to deep extend a default object. Is it ok to use external package like this?

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,158 @@
...
+
+export default class AuFloatingUiModifier extends Modifier {
+  modify(
+    floatingElement,
+    [_referenceElement, _arrowElement],
+    { defaultPlacement = 'bottom-start', options = {}, middleware = [] }
```
brenner-company, line 20: I'm not entirely sure that using `@options` and `@middleware` like this is the correct way to do it?

Regarding `@options`: is it better to split up the individual values to something like `@floaterOffset`, `@arrowOffset`, etc?
Regarding `@middleware`: should there be this much control over the configuration of the actual floating-ui component?

.storybook/preview-body.html
```diff
@@ -51,6 +51,46 @@
   .innerZoomElementWrapper > div > div > div {
     min-height: 300px !important;
   }
+
+  .sb-decorator-example-scroller {
```
brenner-company, line 55: @Dietr Is it possible to place this styling inside a SCSS file that is only used within Storybook? I'm not sure this is the correct place for it.

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,158 @@
...
+      floatingElement instanceof HTMLElement
+    );
+
+    assert(
+      `AuFloatingUI (modifier): The reference and floating elements cannot be the same element.`,
+      floatingElement !== _referenceElement
+    );
+
+    assert(
+      `AuFloatingUI (modifier): @placement must start with either 'bottom' or 'top'.`,
```
brenner-company, line 60: I only provided support for placement of the floater on the `top` and `bottom` side of the reference element. I presume this is enough for now? If needed, this can be easily extended to support `right` and `left`.

stories/5-components/Helpers/AuFloatingUi.stories.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+  parameters: {
+    layout: 'padded',
+  },
+};
+
+const Template = (args) => ({
+  template: hbs`
+  <div class="sb-decorator-example-scroller" data-example-scroller>
+    <div class="sb-decorator-example-scroller-spacer">
```
brenner-company, line 37: @Dietr I presume this can be converted to a Storybook add-on? Can I give it a go or do you already have some experience in this area?

In addition to this set-up, I would also like to add some functionality to automatically center the reference element within the container. Something like this:

```js
 const scroller = document.querySelector('[data-example-scroller]');

  if (scroller) {
    const child = scroller.children[0];

    const scrollerDimensions = {
      width: scroller.offsetWidth,
      height: scroller.offsetHeight,
    };

    const childDimensions = {
      width: child.offsetWidth,
      height: child.offsetHeight,
    };

    if (childDimensions.width > scrollerDimensions.width) {
      scroller.scrollLeft = (childDimensions.width - scrollerDimensions.width) / 2;
    }

    if (childDimensions.height > scrollerDimensions.height) {
      scroller.scrollTop = (childDimensions.height - scrollerDimensions.height) / 2;
    }
  }
```

.storybook/preview-body.html
Dietr: So far I have not found a better way of handling css overrides in storybook.

Not sure what this does exactly but maybe it's also possible to do in the ember test application? There it's easier to manage styles.

stories/5-components/Helpers/AuFloatingUi.stories.js
Dietr: I don't have experience with writing storybook addons. I kind of bypass this by writing demo components in the ember test application.

addon/modifiers/au-floating-ui.js
Windvis: Sure, as long as it's listed as a `dependency` it should work in consuming apps as well.

stories/5-components/Helpers/AuFloatingUi.stories.js
brenner-company: Thanks for the advice! I've put the extracted functionality inside: 
- tests/dummy/app/components/au-example-scroller.hbs
- tests/dummy/app/components/au-example-scroller.js

brenner-company:
@Dietr @Windvis I think this one is good to go 👍 Even did a small test within the `auDropdown` component and it seems to work well.

Windvis:
(Also, FYI, I don't think Dieter is working on Appuniversum anymore)

brenner-company:
> Nice work 👍!

Thanks!

> For now I think it would be better to keep this functionality "private", so only available for internal use in Appuniversum. If we want to use the functionality in apps as well I think a dedicated generic floating-ui addon would be better suited.
> 
> If we go for the private variant I think it makes sense to create an `addon/private` folder where we can store things. Since we now support only 3.28+ we can remove the app reexports and simply import the private component / modifier where needed without it being available in the "global" template scope.

Sure thing!

Regarding Storybook, I would rather like to keep that story so that Chromatic can run visual tests on it. What would be the best way to go (knowing the component is for internal usage only): keep the story visible in the sidebar but under something like 'internal usage only' or hide the story using a CSS solution mentioned here: https://github.com/storybookjs/storybook/issues/9209 (an internal solution to hide stories in the sidebar is not implemented yet sadly). What do you think?

addon/modifiers/au-floating-ui.js
brenner-company: I don't have much experience regarding modifiers, but this seems like a correct conversion, right?

https://github.com/appuniversum/ember-appuniversum/pull/344/commits/0526e4b4c230acb88000c64f57361b273adf715c

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,159 @@
...
+  computePosition,
+  flip,
+  hide,
+  offset,
+  arrow,
+} from '@floating-ui/dom';
+
+import { merge } from 'merge-anything';
+
+export default class AuFloatingUiModifier extends Modifier {
```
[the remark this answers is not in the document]
brenner-company: I don't have much experience regarding modifiers, but this seems like a correct conversion, right?

https://github.com/appuniversum/ember-appuniversum/pull/344/commits/0526e4b4c230acb88000c64f57361b273adf715c

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,161 @@
...
+        Object.assign(arrowElement.style, {
+          [mainSide]: `${-options.arrow.offset}px`,
+        });
+      }
+    };
+
+    let cleanup = autoUpdate(referenceElement, floatingElement, update);
+
+    return () => {
+      cleanup();
```
brenner-company, line 157: @Windvis I noticed something weird when testing the cleanup function: when utilising the floating-ui abilities as a modifier, the cleanup (destroyer) function gets called as you would expect. But when I use the `AuFloatingUi` component this doesn't happen for a reason (maybe it doesn't have to?). Do you have any experience regarding modifiers? 🤞

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,161 @@
+import { modifier } from 'ember-modifier';
```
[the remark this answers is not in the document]
brenner-company: @NullVoxPopuli See my comment at https://discord.com/channels/480462759797063690/1054380851665834046/1054414088769376256. But would certainly would like to collaborate on that 👍

addon/modifiers/au-floating-ui.js
Windvis: I would expect that the component version works the same. How did you test it exactly? Maybe it's a bug in the `modifier` helper?

addon/modifiers/au-floating-ui.js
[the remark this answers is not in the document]
Windvis: Oh, I didn't know about ember-velcro. Using that instead of a custom version seems better indeed 👍.

addon/modifiers/au-floating-ui.js
brenner-company: I was thinking in the same direction, so I made a bug report within `ember-modifier` at https://github.com/ember-modifier/ember-modifier/issues/613 the previous week. Fingers crossed it is related to that helper 🤞

addon/modifiers/au-floating-ui.js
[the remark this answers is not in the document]
brenner-company: Initially, I was thinking the same thing, but when implementing, I noticed the following:
- There were some functional limitations (and bugs) to `ember-velcro`, where as with `ember-appuniversum` we need more customisation.
- At the moment there is no support for the `arrow` element (from `floating-ui`) within `ember-velcro`.
- I'm not the biggest fan of adding an extra layer between our code and `floating-ui`, when it's just a little bit of effort of adding it ourselves.

Also a small remark that the issue noted at https://github.com/appuniversum/ember-appuniversum/pull/344#discussion_r1050653088 also persists within `ember-velcro`. If I can find a solution to that, I'll surely implement it, together with any other improvements, to `ember-velcro` 👍

Does that seem logical?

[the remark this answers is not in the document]
Windvis: Fair enough, I didn't take a detailed look and assumed ember-velcro was the more generic / flexible version 👍.

addon/modifiers/au-floating-ui.js
Windvis: Since this is for internal use only, can't we simply drop the component version for now and use the modifier directly? That would unblock your dropdown issue at least I think?

brenner-company: @Windvis I'll give this a go tomorrow and let you know if this a possible solution 👍

brenner-company: I gave it a go within #349 (as I have an ideal test case already there). Seems to work fine just as a modifier.

Does these changes look ok for you? https://github.com/appuniversum/ember-appuniversum/pull/349/commits/fd16e8b89a3869bd114051ffed274e881c75291b

app/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1 @@
+export { default } from '@appuniversum/ember-appuniversum/modifiers/au-floating-ui';
```
Windvis, line 1: I would remove this re-export and just import the modifier where needed.

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,161 @@
...
+    );
+
+    assert(
+      `AuFloatingUI (modifier): No floating element was defined.`,
+      floatingElement instanceof HTMLElement
+    );
+
+    assert(
+      `AuFloatingUI (modifier): The reference and floating elements cannot be the same element.`,
+      floatingElement !== _referenceElement
```
Windvis, line 55:
```suggestion
      floatingElement !== referenceElement
```

addon/modifiers/au-floating-ui.js
```diff
@@ -0,0 +1,161 @@
...
+
+    assert(
+      `AuFloatingUI (modifier): No reference element was defined.`,
+      referenceElement instanceof HTMLElement
+    );
+
+    assert(
+      `AuFloatingUI (modifier): No floating element was defined.`,
+      floatingElement instanceof HTMLElement
+    );
```
Windvis, line 51: I don't think this check is needed since invoking a modifier guarantees it?

Windvis:
Also, since the component version has a bug, it probably makes sense to not include it for now? I would push a new branch with the current state and remove the component version + docs from this one. That way we can pick it back up when the issue is resolved?

addon/modifiers/au-floating-ui.js
Windvis: Left some more notes, other than that I think it's good to go if everything works as expected.

brenner-company:
> I would push a new branch with the current state and remove the component version + docs from this one. That way we can pick it back up when the issue is resolved?

Done! Those changes will be available inside the `floating-ui-helpers-with-component-variant` branch.

brenner-company:
@Windvis I think this one is good to go 🤞 Feel free to give the final ok. (I'll finalise the integration of the floating-ui helper within the dropdown component when this is done)

tests/dummy/app/components/au-example-scroller.hbs
```diff
@@ -0,0 +1,5 @@
+<div {{this.exampleScrollerModifier}} class="au-d-example-scroller">
```
Windvis, line 1: I don't think this is used / needed anymore now?

brenner-company: I'm using this to test the positioning of the floating element (when using floating-ui), but I can remove it if you like and keep a copy of it locally 👍

PR #345, opened 2022-11-21
Addition of support for block content within checkbox component

brenner-company:
The addition of support for block content within the `au-control-checkbox` component.

The reason for this is mostly cases like this within Kaleidos:

CleanShot 2022-11-21 at 11 25 13

### DIFF
--- a/addon/components/au-control-checkbox.hbs
+++ b/addon/components/au-control-checkbox.hbs
@@ -13,5 +13,9 @@
     ...attributes
   />
   <span class="au-c-control__indicator"></span>
-  {{@label}}
+  {{#if (has-block)}}
+    {{yield}}
+  {{else}}
+    {{@label}}
+  {{/if}}
 </label>
--- a/tests/integration/components/au-control-checkbox-test.js
+++ b/tests/integration/components/au-control-checkbox-test.js
@@ -19,6 +19,16 @@ module('Integration | Component | au-control-checkbox', function (hooks) {
     assert.dom().hasText('foo');
   });
 
+  test('it can show block content', async function (assert) {
+    await render(hbs`<AuControlCheckbox>bar</AuControlCheckbox>`);
+    assert.dom().hasText('bar');
+  });
+
+  test('it will give preference to show block content against a label', async function (assert) {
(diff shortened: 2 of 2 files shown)


Dietr (approved):

Windvis (commented):
Nice 👍. Would it make sense to deprecate the `@label` version?

It seems this component already [has some tests](https://github.com/appuniversum/ember-appuniversum/blob/506fd1e559021ef2b759977f6b0a6604ea9f85e3/tests/integration/components/au-control-checkbox-test.js), could you add some for this behavior as well?

A variant of [this](https://github.com/appuniversum/ember-appuniversum/blob/506fd1e559021ef2b759977f6b0a6604ea9f85e3/tests/integration/components/au-control-checkbox-test.js#L17-L20) would be nice. And I would also add a test which checks that the value of the block has priority over the `@label` version.

```hbs
<AuControlCheckbox @label="Foo">Bar</AuControlCheckbox>
```
For example this would output `Bar` instead of `Foo` with the current implementation so a test for that expectation would be nice.

brenner-company:
> It seems this component already [has some tests](https://github.com/appuniversum/ember-appuniversum/blob/506fd1e559021ef2b759977f6b0a6604ea9f85e3/tests/integration/components/au-control-checkbox-test.js), could you add some for this behavior as well?
> 
> A variant of [this](https://github.com/appuniversum/ember-appuniversum/blob/506fd1e559021ef2b759977f6b0a6604ea9f85e3/tests/integration/components/au-control-checkbox-test.js#L17-L20) would be nice. And I would also add a test which checks that the value of the block has priority over the `@label` version.
> 
> ```
> <AuControlCheckbox @label="Foo">Bar</AuControlCheckbox>
> ```
> 
> For example this would output `Bar` instead of `Foo` with the current implementation so a test for that expectation would be nice.

Good idea, forgot about that 👍 Done!


> Would it make sense to deprecate the `@label` version?

I'm on the fence on that one to be honest. Do we see the block content as an extension or as a replacement? Inside Kaleidos we have a similar component, but there we don't really document the possibility to use block content, to prevent accidental faulty styling of the label a bit 🤷‍♂️

@Dietr Do you have an opinion on this?

Windvis (approved):

PR #349, opened 2022-12-07
Application of floating UI helpers within dropdown component

### LINKED ISSUE #333: More dynamic display of menu element within dropdown component

brenner-company:
The application of the newly created `AuFloatingUi` component to the `AuDropdown` component.

See #333 for additional information.

⚠️ PR #344 must be merged before finalising this one (as this is based on that specific branch)! Also the base for branch needs to be switched to `development` before merging.

Closes #333


testem.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = {
-  test_page: 'tests/index.html?hidepassed',
+  test_page: 'tests/index.html?hidepassed&fullscreencontainer',
```
brenner-company, line 4: Because of the failing tests for the dropdown component, I had to add this parameter. This will make sure that tests  are being run without any scaling (normally at 50%) so that correctly positioning the dropdown doesn't break.

For more information see: [this](https://github.com/emberjs/ember-qunit/issues/521) (similar issue) & [this](https://github.com/emberjs/ember-qunit/pull/763) (solution).

addon/components/au-dropdown.hbs
```diff
@@ -18,18 +19,30 @@
...
-    role="menu"
-    tabindex="-1"
-    {{focus-trap
-      isActive=this.dropdownOpen
-      shouldSelfFocus=true
-      focusTrapOptions=(hash
-        clickOutsideDeactivates=this.clickOutsideDeactivates
-      )
+    {{
+      au-floating-ui
```
Windvis, line 23: I would import the modifier in the backing class and store it on a property. That aligns better with the future strict mode components and means the modifier is truly for internal use only for now.

brenner-company: just to be sure, is the following code example in line with what you meant?

`au-dropdown.js`
```js
import AuFloatingUiModifier from '@appuniversum/ember-appuniversum/modifiers/au-floating-ui';

export default class AuDropdown extends Component {
  auFloatingUi = AuFloatingUiModifier;

  // etcetera
}
```

`au-dropdown.hbs`
```handlebars
<div
  {{this.auFloatingUi
    this.referenceElement
    this.arrowElement
    defaultPlacement=this.alignment
    options=this.floatingUiOptions
  }}
>
  <!-- etcetera -->
</div>
```

Windvis: Yup! With some small nitpicks. I would only move the file to a "private" folder or something (and drop the `Au` prefix since it's only used internally anyways). So something like this:

```js
import FloatingUiModifier from '@appuniversum/ember-appuniversum/private/modifiers/floating-ui';
```

brenner-company: Oh, I think you mentioned this a while ago. I'll get it done 👍 Thanks!

addon/components/au-dropdown.js
```diff
@@ -52,9 +52,9 @@ export default class AuDropdown extends Component {
 
   // Dropdown alignment
   get alignment() {
-    if (this.args.alignment == 'left') return 'au-c-dropdown__menu--left';
-    if (this.args.alignment == 'right') return 'au-c-dropdown__menu--right';
-    return '';
+    if (this.args.alignment == 'left') return 'bottom-start';
```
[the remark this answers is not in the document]
Windvis: I'm not sure. Are those other values useful for a "dropdown"?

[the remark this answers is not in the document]
brenner-company: I gave this another look and I have to agree that these options seem sufficient. Thanks for the feedback!

brenner-company:
@Windvis I think this one is good to go 👍 Just let me know if you have anymore question or remarks.

testem.js
Windvis: I looked into this a bit since it also made the tests fail on the `/tests`. It seems this was code by a bug in the clipping detection code of floating ui and it seems to be fixed on the latest 1.1.0 release. So I think we can revert this and simply bump the version.

addon/components/au-dropdown.hbs
```diff
@@ -18,18 +19,29 @@
...
-    class="au-c-dropdown__menu {{this.alignment}} {{if this.dropdownOpen 'is-visible'}}"
-    role="menu"
-    tabindex="-1"
-    {{focus-trap
-      isActive=this.dropdownOpen
-      shouldSelfFocus=true
-      focusTrapOptions=(hash
-        clickOutsideDeactivates=this.clickOutsideDeactivates
-      )
+    {{this.floatingUi
```
Windvis, line 22: Looking into the test problem I realized that the floating-ui modifier is always executed, even if the dropdown isn't open, which doesn't sound ideal? 

If we want to fix that I think we have a couple of options:

1. Wrap this div in an `{{#if this.dropdownOpen}}` block. That would simply not output the html until the dropdown is open. Usually this is the prefered option since it means less things have to render / execute. I'm not sure why that wasn't the case in the previous implementation. Could that be due for a11y reasons?
2. We add an `isActive` option to the modifier similar to how the focus-trap modifier works. The modifier would then simply early return if it's set to false and non of the floating-ui code would run.
3. We conditionally apply the modifier:
  ```hbs
    <div {{(if this.dropdownOpen (modifier this.floatingUI .....))}}></div>
  ```
  This will run into the same issue as the component version though, so not really an option at the moment.

addon/components/au-dropdown.hbs
```diff
@@ -18,18 +19,29 @@
...
+      this.referenceElement
+      this.arrowElement
+      defaultPlacement=this.alignment
+      options=this.floatingUiOptions
     }}
-    {{on "click" this.closeDropdown}}
+    class="au-c-dropdown__floater {{if this.dropdownOpen 'is-visible'}}"
```
Windvis, line 28: If we go for the `{{#if this.dropdownOpen}}` wrapper the is-visible class isn't needed anymore.

addon/components/au-dropdown.hbs
```diff
@@ -18,18 +19,29 @@
...
   >
-    {{yield}}
+    <div {{this.arrow}} class="au-c-dropdown__arrow"></div>
+    <div
+      {{focus-trap
+        isActive=this.dropdownOpen
```
Windvis, line 33: This would then also no longer be needed.

addon/components/au-dropdown.js
```diff
@@ -2,8 +2,30 @@ import Component from '@glimmer/component';
...
 
 export default class AuDropdown extends Component {
+  @tracked referenceElement = undefined;
+  @tracked arrowElement = undefined;
+
+  floatingUi = FloatingUiModifier;
+
+  reference = modifier(
+    (element) => {
+      this.referenceElement = element;
```
Windvis, line 17: Since we now have a reference to the button I think we no longer need to query for it here either:

https://github.com/appuniversum/ember-appuniversum/blob/363616290579144ddcaf2760ce50421aa84a3e40/addon/components/au-dropdown.js#L26

addon/components/au-dropdown.hbs
brenner-company: I think you're right. I assumed that, because the floating-ui documentation doesn't mention this as a possible issue and uses a visibility toggle as a functionality example (see https://floating-ui.com/docs/tutorial#functionality), this wouldn't really be an issue. Thanks for noticing! 

Eventually, I went with option 1. Funny, because when I tried to do the same thing for the proof-of-concept dropdown with PopperJS, this didn't work.

testem.js
brenner-company: Good find! Did a quick test and it seems to work just fine 👍

addon/components/au-dropdown.js
brenner-company: Does this work when you have multiple `AuDropdown` instances on the same page (it will always select the first match)? Or is the query only executed within the scope of the component?

Windvis: No, I think you are right. I should have added some sort of unique-id to the selector I think. Good we can replace it with the modifier now 😄. Not sure why I didn't do it like that in the first place.

Windvis (approved):

PR #357, opened 2023-01-16
Au radio control checked should be without the @

lagartoverde:
Found a bug that the AuRadioButton pass down `checked` instead of `@checked` so I fixed the docs.


Windvis:
Thanks 👍

PR #362, opened 2023-01-31
Add an `@onClose` argument to the `AuDropdown` component 

### LINKED ISSUE #348: Add hook to `<AuDropdown>` to execute action when dropdown closes
We have a dropdown whose content changes if a user selects one of its items. When the dropdown closes (e.g. because the user clicks outside of the dropdown content) and the user re-opens the dropdown, we still render the second set of items, not the original items. Ideally when the dropdown closes, we would reset its content to the original items.

If `<AuDropdown>` had a hook that gets called whenever it gets closed, we could pass an action that resets the boolean that determines which set of items is shown in the dropdown. This way, the next time the user opens the dropdown we will show the original content again.

This specific [dropdown in Kaleidos](https://github.com/kanselarij-vlaanderen/frontend-kaleidos/blob/development/app/components/m-header.hbs#L104-L175).

Windvis:
Closes #348

### DIFF
--- a/addon/components/au-dropdown.js
+++ b/addon/components/au-dropdown.js
@@ -36,11 +36,16 @@ export default class AuDropdown extends Component {
   @action
   closeDropdown() {
     this.dropdownOpen = false;
+    this.args.onClose?.();
   }
 
   @action
   toggleDropdown() {
     this.dropdownOpen = !this.dropdownOpen;
+
+    if (!this.dropdownOpen) {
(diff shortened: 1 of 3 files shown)


Windvis:
@sergiofenoll does this look ok to you?

sergiofenoll:
> @sergiofenoll does this look ok to you?

Perfect :ok_hand: 
Thanks for adding it! :D

PR #363, opened 2023-01-31
Add an `@onClose` argument to the `AuAlert` component

### LINKED ISSUE #348: Add hook to `<AuDropdown>` to execute action when dropdown closes

### LINKED ISSUE #350: Add hook to `AuAlert` that triggers after its closed
The `<AuAlert>` component can be made `closable`. If a closable `<AuAlert>` is conditionally rendered, there's no way to reflect the fact that it has been closed back in the "calling" component, so there's no way to reset the condition that triggers it being rendered.

Add a `didClose` hook to the `<AuAlert>` component that gets called when a user closes an alert. This way callers can pass in an action to reset any state that's used when conditionally rendering the alert.

Windvis:
Closes #350

### DIFF
--- a/addon/components/au-alert.js
+++ b/addon/components/au-alert.js
@@ -22,10 +22,10 @@ export default class AuAlertComponent extends Component {
     return '';
   }
 
-  // @TODO: add icon switching functionality
-
   @action
   closeAlert() {
-    this.isVisible = !this.isVisible;
+    this.isVisible = false;
+
+    this.args.onClose?.();
   }
 }
--- a/stories/5-components/Notifications/AuAlert.stories.js
+++ b/stories/5-components/Notifications/AuAlert.stories.js
@@ -21,6 +21,10 @@ export default {
       description: 'Set the size of the alert',
     },
     closable: { control: 'boolean', description: 'Makes the alert closable' },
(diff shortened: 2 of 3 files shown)


Windvis:
@sergiofenoll Same question here 😄. I went with `onClose` instead of `didClose` to make it more consistent with other action names, but I can change it if you have want.

sergiofenoll:
> @sergiofenoll Same question here smile. I went with `onClose` instead of `didClose` to make it more consistent with other action names, but I can change it if you have want.

`onClose` is good, thank you!

PR #369, opened 2023-02-06
Add a `@maskOptions` argument to the AuInput component 

### LINKED ISSUE #207: Add support for more complex inputmasks
The input mask feature that was recently added to the `AuInput` compontent is great for the simple use cases where only a mask is provided. Trying to use the date extension doesn't seem to work yet though. I haven't dug into it yet but it's related to the way we pass arguments to the InputMask component.

Windvis:
This allows users to pass options directly into the `Inputmask` instance so it's now a lot more flexible.

At its most basic form, `@maskOptions` can be used as a replacement of `@mask` and `@maskPlaceholder`:
```diff
<AuInput
-  @mask="99.99.99-999.99"
-  @maskPlaceholder="_"
+  @maskOptions={{hash
+    mask="99.99.99-999.99"
+    placeholder="_"
+  }}
  @onChange={{this.setRijksregisternummer}}
/>
```

But it allows users to pass any of the [inputmask options](https://github.com/RobinHerbots/Inputmask#options). This opens up the use of aliases, or any of the event callbacks to have more finegrained control.

Closes #207

### DIFF
--- a/addon/components/au-input.js
+++ b/addon/components/au-input.js
@@ -1,15 +1,17 @@
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+import { modifier } from 'ember-modifier';
+import Inputmask from 'inputmask';
 
 export default class AuInput extends Component {
   constructor() {
     super(...arguments);
 
     assert(
-      '<AuInput>: An `@onChange` handler was provided but that will only be called if a `@mask` is provided as well.',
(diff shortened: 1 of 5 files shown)


addon/components/au-input.js
```diff
@@ -43,8 +45,66 @@ export default class AuInput extends Component {
     return this.args.type || 'text';
   }
 
+  get classes() {
+    return [
+      'au-c-input',
+      this.isMasked ? 'au-c-input--mask' : '',
+      this.error,
+      this.warning,
+      this.disabled,
```
Windvis, line 54: I forgot to add the "width" class, good catch Chromatic 👍.

PR #377, opened 2023-02-16
fix(toggle-switch): make width calculations more consistent

abeforgit:
Issue that triggered this:
when the toggle component is a child of a flexbox where there's another element taking up all the leftover width, it doesn't reserve enough space for the toggle to look proper, which did not improve with increasing the width variable.
untoggled(too small):
image
toggled (too far to the right):
image

I don't really know if this is the right solution, but it makes sense to me. It also eliminates the absolute offset in the translation, which caused it to render weird with large width settings:

before the change, with width set to 5rem:
image

after the change, with width set to 5rem:
image

### DIFF
--- a/app/styles/ember-appuniversum/_c-toggle-switch.scss
+++ b/app/styles/ember-appuniversum/_c-toggle-switch.scss
@@ -29,7 +29,7 @@ $au-toggle-switch-background-color-on-disabled   : var(--au-gray-400) !default;
   position: relative;
   cursor: pointer;
   height: $au-toggle-switch-height - .4rem;
-  width: $au-toggle-switch-width * 1.5;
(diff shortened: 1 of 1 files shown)


Windvis:
I'm not sure, but can't we simply tell that the toggle itself is not allowed to shrink (by setting flex-shrink to 0)?

brenner-company:
> I'm not sure, but can't we simply tell that the toggle itself is not allowed to shrink (by setting flex-shrink to 0)?
> 
> @brenner-company Any insights? 😄

I agree with @Windvis on this issue. Disabling `flex-shrink` is the most elegant solution here 👍

Windvis:
Closing since #380 fixes the issue in a different way. Let us know if the issue still persists on v2.4.2.

PR #378, opened 2023-02-16
build(dev): set isDevelopingAddon when appropriate

abeforgit:
Kind of incredible this isn't the default in Ember, but this sets the isDevelopingAddon to true when running the host app in dev mode, which means that the host app will rebuild when you change something in the npm-linked local copy of ember-appuniversum

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,9 @@
 
 module.exports = {
   name: require('./package').name,
+  isDevelopingAddon() {
+    return process.env.EMBER_ENV === 'development';
+  },
   options: {
     babel: {
       // This is needed for dynamic imports to work: https://github.com/ef4/ember-auto-import#installing-ember-auto-import-in-an-addon


Windvis:
It's not the default because it would mean all addons would be watched when the app is in dev mode which I assume has impact on system resources? I also don't think addons should decide when they need to be watched. Developing addons by linking them to apps isn't a good practice either. It's very easy to accidentally depend on behavior that is not safe to depend on in addons (prototype extensions, dependencies that aren't defined in the addon, ..). The test app / storybook should be used instead and you can then still do a sanity check afterwards by linking it into a real app.

That hook is also being dropped in v2 addons:
> isDevelopingAddon: Dropped. This doesn’t belong in each addon’s code, it’s a runtime decision and control over it belongs in ember-cli proper. Under embroider developers can set a comma-separated list of addon package names in the EMBROIDER_REBUILD_ADDONS environment variable instead.

So I would rather not introduce this change now.

PR #379, opened 2023-02-20
Only call `.focus` if the element exist

### LINKED ISSUE #376: The `AuMainHeader` component expects an element with id `main` to be present.
**Description**
The `AuMainHeader` component expects an element with id `main` to be present (such as an `AuMainContainer`). It's possible that this is not always the case. If an element with id `main` is not present, the component throws an error.

Windvis:
This fixes an issue where the AuMainHeader component would trigger a type error exception if the `#main` element didn't exist.

Closes #376

### DIFF
--- a/addon/components/au-main-header.js
+++ b/addon/components/au-main-header.js
@@ -4,7 +4,6 @@ import { action } from '@ember/object';
 export default class AuMainHeader extends Component {
   @action
   headerLinkFocus() {
-    // Focus content window
-    document.querySelectorAll('#main')[0].focus();
+    document.querySelector('#main')?.focus();
   }
 }


addon/components/au-main-header.js
```diff
@@ -4,7 +4,6 @@ import { action } from '@ember/object';
 export default class AuMainHeader extends Component {
   @action
   headerLinkFocus() {
-    // Focus content window
-    document.querySelectorAll('#main')[0].focus();
+    document.querySelector('#main')?.focus();
```
Windvis, line 7: Should we display a warning in development builds? I think this setup is here for accessibility reasons.

PR #380, opened 2023-02-20
Fix a `<AuToggleSwitch>` shrinking issue

Windvis:
This change prevents the visual switch from shrinking below its minimum width.

Before:
image

After:
image

Supersedes #377


abeforgit:
Not to be pedantic, but elements can't shrink below their minimum width anyway (or at least not in a flex context). The toggle just doesn't set any min-width at all. My original solution was also wrong (cause I misremembered how `width`, `min-width` and `max-width` interact), but it seems weird to me to disable any and all shrinking rather than just setting a minimum width

Issue #367, opened 2023-02-06
Add `AuCheckbox` and `AuRadio` components


Windvis:
We already have the `AuControlCheckbox` and `AuControlRadio` components, but there are some inconsistencies between the 2. We should re-evaluate the implementations of both components and create new ones under the more intuitive names. This has the benefit that apps need to "opt-in" into using the new components and this isn't a breaking change.

Relates to: https://github.com/appuniversum/ember-appuniversum/issues/365#issuecomment-1418722020

Issue #356, opened 2023-01-13
Addition of radio list component


brenner-company:
Within Kaleidos we have the possibility to use a radio list component that provides the following functionality: automatically group radio components together by name (by generating one with something like: `radio-${guidFor(this)}`) so that you don't have to do it yourself by providing a custom (unique) one.

Is this something that could be interesting to integrate within `ember-appuniversum`?

A screenshot a similar radio list component (inside the Kaleidos styleguide) can be found below:

<img width="598" alt="CleanShot 2023-01-13 at 13 45 22@2x" src="https://user-images.githubusercontent.com/18174827/212323435-e96b102c-37fa-4173-927e-a46151f02066.png">

Windvis:
I think it makes sense, yes. If we have dedicated checkbox / radio group components I think it also makes sense to let them handle the checked and disabled states of their inputs, and maybe the change event as well?

Something similar to this component: https://velvet-thunder.netlify.app/docs/components/radio-group

PR #381, opened 2023-02-22
Addition of new radio, checkbox and group components

brenner-company:
- The addition of new radio & checkbox components alongside the legacy versions of those, named `AuControlRadio` & `AuControlCheckbox`. This adjustment should close https://github.com/appuniversum/ember-appuniversum/issues/367.
- The addition of new radio-group & checkbox-group components. This adjustment should close https://github.com/appuniversum/ember-appuniversum/issues/356.

### DIFF
--- a/addon/components/au-radio.js
+++ b/addon/components/au-radio.js
@@ -1,0 +1,29 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class AuRadioComponent extends Component {
+  get checked() {
+    const { inGroup, checked, groupValue, value } = this.args;
+    return inGroup ? value === groupValue : checked;
+  }
+
+  get value() {
+    const { value } = this.args;
+    return value === undefined ? undefined : String(value);
+  }
+
+  @action
+  onChange(event) {
+    if (this.args.disabled === true) {
+      return;
+    }
+
+    const { inGroup, onChange, onChangeGroup, value } = this.args;
+
+    if (inGroup && typeof onChangeGroup === 'function') {
+      onChangeGroup(value, event);
+    } else if (typeof onChange === 'function') {
+      onChange(event.target.value, event);
+    }
+  }
+}
--- a/tests/integration/components/au-radio-group-test.js
+++ b/tests/integration/components/au-radio-group-test.js
@@ -1,0 +1,87 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'dummy/tests/helpers';
+import { click, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | au-radio-group', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it can be given a name', async function (assert) {
+    await render(hbs`
+      <AuRadioGroup @name="foo" as |Group|>
+        <Group.Radio data-test-foo>Foo</Group.Radio>
+      </AuRadioGroup>
+    `);
+    assert.dom('[data-test-foo]').hasAttribute('name', 'foo');
+  });
+
+  test('if it is given no name, it will automatically generate a unique id', async function (assert) {
+    await render(hbs`
+      <AuRadioGroup as |Group|>
+        <Group.Radio data-test-foo>Foo</Group.Radio>
+      </AuRadioGroup>
+    `);
+    assert.dom('[data-test-foo]').hasAttribute('name', /^ember\d+/); // E.g. 'ember308'
+  });
+
+  test('it can be given a value', async function (assert) {
+    this.groupValue = 'bar';
+    await render(hbs`
+      <AuRadioGroup @value={{this.groupValue}} as |Group|>
+        <Group.Radio @value="foo" data-test-foo>Foo</Group.Radio>
+        <Group.Radio @value="bar" data-test-bar>Bar</Group.Radio>
+      </AuRadioGroup>
+    `);
+
+    assert.dom('[data-test-bar]').isChecked();
+
+    this.set('groupValue', 'foo');
+    assert.dom('[data-test-foo]').isChecked();
+  });
+
+  test('it can be disabled', async function (assert) {
+    this.disabled = true;
+    await render(hbs`
+      <AuRadioGroup @disabled={{this.disabled}} as |Group|>
+        <Group.Radio data-test-foo>Foo</Group.Radio>
+      </AuRadioGroup>
+    `);
+
+    assert.dom('[data-test-foo]').isDisabled();
+
+    this.set('disabled', false);
+    assert.dom('[data-test-foo]').isNotDisabled();
+  });
+
+  test('it calls the provided @onChange action when its state changes by user input', async function (assert) {
+    assert.expect(4);
+
+    this.value = null;
+    this.handleChange = (value, event) => {
+      this.set('value', value);
+      assert.true(event instanceof Event);
+    };
+
+    await render(hbs`
+      <AuRadioGroup @onChange={{this.handleChange}} as |Group|>
+        <Group.Radio @value="foo" data-test-foo>Foo</Group.Radio>
+        <Group.Radio @value="bar" data-test-bar>Bar</Group.Radio>
+      </AuRadioGroup>
+    `);
+
+    await click('[data-test-bar]');
+    assert.strictEqual(this.value, 'bar')
+
+    await click('[data-test-foo]');
+    assert.strictEqual(this.value, 'foo')
+  });
+
+  test('it adds any extra attributes to the group element', async function (assert) {
+    await render(hbs`
+      <AuRadioGroup foo="bar" as |Group|>
+        <Group.Radio>Foo</Group.Radio>
+      </AuRadioGroup>
+    `);
+    assert.dom('[data-test-radio-group]').hasAttribute('foo', 'bar');
+  });
+});
--- a/addon/components/au-radio-group.hbs
+++ b/addon/components/au-radio-group.hbs
@@ -1,0 +1,19 @@
+<div
+  class="au-c-control-group {{this.alignmentClass}}"
+  role="group"
+  ...attributes
+  data-test-radio-group
+>
+  {{yield
+    (hash
+      Radio=(component
+        "au-radio"
+        name=(or @name this.uniqueName)
+        groupValue=@value
+        disabled=@disabled
+        onChangeGroup=@onChange
+        inGroup=true
+      )
+    )
+  }}
+</div>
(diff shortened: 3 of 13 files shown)


brenner-company:
At the moment, I only added the new `AuRadio` & `AuRadioGroup` components. If these two check all the boxes, I'll do the same thing for `AuCheckbox` & `AuCheckboxGroup`.

brenner-company:
A few remarks/questions for whoever wants to be involved:
- The new form components won't have any connection anymore with the `control` name(space), but there are still a few references regarding the shared (S)CSS/styling (`au-c-control`, etc.). Should this also be refactored (mostly regarding name)? If yes, should this be split up (radio + checkbox)? I'm not entirely sure this would be worth the effort at the moment.
- As I copied a few things from https://velvet-thunder.netlify.app/docs/components/radio-group (as @Windvis mentioned inside https://github.com/appuniversum/ember-appuniversum/issues/356), I noticed the following: inside this piece of code there is a check wether a `onChange` function is defined either on the group component or the radio component itself (see [reference](https://github.com/Bagaar/velvet-thunder/blob/90935a88beb5c6c71e3090e64d4b6ed5c1b92fbf/addon/components/velvet-radio/index.ts#L47)). In case of the later, the `event.target.checked` value is returned. I changed this to the `value` in our version (as we did within `AuControlRadio`). Does this seem more logical?

Windvis:
(FYI I'll try to take a look at this ASAP, but I'm a bit busy this sprint so it will probably be late tomorrow or early next week)

brenner-company:
@Windvis No hurry ☺️ I'll try and finalise the last remaining component (checkbox group) tomorrow morning.

brenner-company:
Some more remarks/questions (regarding checkbox & checkbox-group):
- A similar question as for `AuRadioGroup`: I copied a few things from https://velvet-thunder.netlify.app/docs/components/checkbox-group but made some adjustments on my own: I removed the possibility to pass the (group)value as an object as I thought that doing this with an array is just as sufficient & it removes some unnecessary complexity (see [reference](https://github.com/Bagaar/velvet-thunder/blob/90935a88beb5c6c71e3090e64d4b6ed5c1b92fbf/addon/components/velvet-checkbox/index.ts#L32)). Does this seem logical?
- I assume it will be up to the developer to use a tracked variable for `AuCheckboxGroup` (and also `AuRadioGroup` I guess?) for `@value` and when a callback function is called, reassign the new value(s) back to that tracked variable?

brenner-company:
@Windvis Do you have, by any chance, a timeframe for this? Much thanks!

addon/components/au-checkbox.js
```diff
@@ -0,0 +1,45 @@
...
+  get checked() {
+    const { inGroup, checked } = this.args;
+    return inGroup ? this.isCheckedInGroup : checked;
+  }
+
+  @action
+  onChange(event) {
+    if (this.args.isDisabled === true) {
+      return;
+    }
```
Windvis, line 23: I'm not sure if this is needed. I don't think browsers trigger the event if the input is disabled?

Also, I think it needs to be `this.args.disabled` in our version and we can also remove the `=== true`?

```suggestion
    if (this.args.disabled) {
      return;
    }
```

tests/integration/components/au-radio-group-test.js
```diff
@@ -0,0 +1,87 @@
...
+    assert.dom('[data-test-foo]').hasAttribute('name', 'foo');
+  });
+
+  test('if it is given no name, it will automatically generate a unique id', async function (assert) {
+    await render(hbs`
+      <AuRadioGroup as |Group|>
+        <Group.Radio data-test-foo>Foo</Group.Radio>
+      </AuRadioGroup>
+    `);
+    assert.dom('[data-test-foo]').hasAttribute('name', /^ember\d+/); // E.g. 'ember308'
```
Windvis, line 24: I wouldn't test it like that since the output an implementation detail. I would just check if the attribute exists since that means _some_ value is set.

A better way could be to check multiple radios and test if only 1 of them stays checked?

tests/integration/components/au-radio-group-test.js
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  test('it can be given a value', async function (assert) {
+    this.groupValue = 'bar';
+    await render(hbs`
+      <AuRadioGroup @value={{this.groupValue}} as |Group|>
```
Windvis, line 30: I wonder if we should rename `@value` to `@selected` similar to ember-power-select? It looks a bit strange since the radios also use the `@value` argument.

addon/components/au-checkbox-group.hbs
```diff
@@ -0,0 +1,18 @@
...
+  data-test-checkbox-group
+>
+  {{yield
+    (hash
+      Checkbox=(component
+        "au-checkbox"
+        inGroup=true
+        groupValue=@value
+        disabled=@disabled
+        onChangeGroup=@onChange
```
Windvis, line 14: Hmm, I'm not sure if I like this. The checkbox component now has to know about implementation details of the group.

I think it might be better to create a new private wrapper component that we yield instead. That component could contain the logic and then pass the normal `@onChange` and `@checked` arguments to the checkbox component. It's an extra layer but I think it will make the implementation more clear.

I poked at it in a separate PR and it seems to work.

It doesn't have to block the merging of this Pr though, we can change the implementation in a future PR.

addon/components/au-radio.js
```diff
@@ -0,0 +1,29 @@
...
+export default class AuRadioComponent extends Component {
+  get checked() {
+    const { inGroup, checked, groupValue, value } = this.args;
+    return inGroup ? value === groupValue : checked;
+  }
+
+  get value() {
+    const { value } = this.args;
+    return value === undefined ? undefined : String(value);
+  }
```
Windvis, line 13: I know it's implemented like this in the velvet version, but we don't need this here since we don't use TS / Glint. I would just pass through the argument to the attribute directly.

stories/5-components/Forms/AuCheckboxGroup.stories.js
```diff
@@ -0,0 +1,52 @@
...
+      control: 'array',
+      description:
+        'Used to identify which checkboxes in the group are selected.',
+    },
+    disabled: {
+      control: 'boolean',
+      description: 'Adds a disabled state to all the checkboxes.',
+    },
+    onChange: {
+      control: 'function',
```
Windvis, line 21: Storybook has an actions system that can be used to log messages in the docs. I think we do it in a couple of places already.

```suggestion
      action: 'change', // or whatever name you choose
```

stories/5-components/Forms/AuCheckboxGroup.stories.js
```diff
@@ -0,0 +1,52 @@
...
+    </AuCheckboxGroup>`,
+  context: args,
+});
+
+export const Component = Template.bind({});
+Component.args = {
+  alignment: 'default',
+  value: ['checkboxTwo', 'checkboxThree'],
+  disabled: false,
+  onChange: '',
```
Windvis, line 51:
```suggestion
```

Windvis (commented):
Thanks for working on this and sorry for the long delay. I don't really have reserved time for Appuniversum work so I do it when possible, but the larger the PR the less likely I can squeeze it in :smile:.

Now that I look at this in some more detail I'm not 100% convinced by the API that the velvet implementation used. I think something like power-select might have been more interesting.

Anyways, it probably doesn't matter that much, so we can just go with this version and revise in the future if needed.

I left some minor remarks, but non of them are blockers for merging, so feel free to merge and release if you want.

tests/integration/components/au-checkbox-group-test.js
```diff
@@ -0,0 +1,69 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | au-checkbox-group', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it can be given a value', async function (assert) {
+    this.groupValue = ['bar'];
+    await render(hbs`
+      <AuCheckboxGroup @value={{this.groupValue}} as |Group|>
+        <Group.Checkbox @name="foo" data-test-foo>Foo</Group.Checkbox>
```
Windvis, line 13: Should we also go for `@value` here instead of `@name`?

Windvis:
(I pushed a commit to master that should fix the ember-qunit peerdep issue so you can rebase / merge to resolve that if you want)

addon/components/au-checkbox.js
brenner-company: I saw this within the Velvet implementation but I was also in doubt this should be added/copied. Removed ✅

tests/integration/components/au-radio-group-test.js
brenner-company: Sounds more logical to me 👍 I've adjusted the existing test & extended the one after that to cover your last question.

tests/integration/components/au-radio-group-test.js
brenner-company: I think I took a peek around inside some other frameworks and it seemed like this was largely the default name.

Also, if you need get the selected value of a list in vanilla JS, you'll need something like this: `document.querySelector('input[name="foo"]:checked').value`.

But I'm still a bit on the fence regarding that. If you think it will bring some extra clarity, sure, why not.

brenner-company: Maybe I'll ask around within the Kaleidos dev team on Wednesday (Tuesday is my 'day off' every week btw).

addon/components/au-checkbox-group.hbs
brenner-company: Seems ok to me 👍 Will the structure (in terms of usage) change with this PR?

tests/integration/components/au-checkbox-group-test.js
brenner-company: > Should there perhaps also be an option for a group name? For the sake of HTML semantics? (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes as example)

I found the following inside the story for `AuCheckbox`: `name: { control: 'text', description: 'Groups checkboxes for server side' }`, this seems like something I copied from the old checkbox component but I sure seems logical, right?

tests/integration/components/au-radio-group-test.js
brenner-company: > Maybe I'll ask around within the Kaleidos dev team on Wednesday (Tuesday is my 'day off' every week btw).

`@selected` it is 👍 

Changed it for both checkbox- and radio-group components.

tests/integration/components/au-checkbox-group-test.js
brenner-company: I've also added `@name` to the checkbox & checkbox-group component, but, for the latter, only as a custom value, not an automatic one (as we do for radio-group).

Windvis (approved):

PR #390, opened 2023-03-17
Refactor the control group components

Windvis:
This allows us to split off the group logic from the core checkbox logic, which makes the component easier to maintain.

### DIFF
--- a/addon/components/au-checkbox-group.hbs
+++ b/addon/components/au-checkbox-group.hbs
@@ -7,8 +7,7 @@
   {{yield
     (hash
       Checkbox=(component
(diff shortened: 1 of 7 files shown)


addon/components/au-checkbox-group.hbs
```diff
@@ -7,8 +7,7 @@
   {{yield
     (hash
       Checkbox=(component
-        "au-checkbox"
-        inGroup=true
+        this.GroupCheckbox
```
Windvis, line 10: We still need to use ensureSafeComponent here.

PR #391, opened 2023-03-17
Patching of ember-focus-trap and application to modal component

brenner-company:
Local patching of `ember-focus-trap` (regarding https://github.com/josemarluedke/ember-focus-trap/issues/71) & application of said changes to `AuModal` component.

PR #395, opened 2023-04-24
Update of ember-focus-trap package and application of additional elements argument to modal component

brenner-company:
This PR continues the works already done within https://github.com/appuniversum/ember-appuniversum/pull/391 but without the patching of the `ember-focus-trap` package:

- Update of the `ember-focus-trap` package (new release contains https://github.com/josemarluedke/ember-focus-trap/issues/71).
- Addition of `additionalElements` argument (with `#ember-basic-dropdown-wormhole` as a default) to the `AuModal` component to enable focus/tabbing within `ember-power-select` instances when following variables are met:
  - The `AuModal` instance has `@overflow` disabled.
  - The `ember-focus-trap` instance has `@renderInPlace` disabled.

### DIFF
--- a/addon/components/au-modal.hbs
+++ b/addon/components/au-modal.hbs
@@ -20,6 +20,7 @@
       tabindex="-1"
       {{focus-trap
         isActive=@modalOpen
+        additionalElements=(array '#ember-basic-dropdown-wormhole')
         focusTrapOptions=(hash
           initialFocus=this.initialFocus
           allowOutsideClick=true
--- a/addon/components/au-modal.hbs
+++ b/addon/components/au-modal.hbs
@@ -20,7 +20,7 @@
       tabindex="-1"
       {{focus-trap
         isActive=@modalOpen
-        additionalElements=(array "#ember-basic-dropdown-wormhole")
+        additionalElements=this.additionalElements
         focusTrapOptions=(hash
           initialFocus=this.initialFocus
           allowOutsideClick=true
--- a/addon/components/au-modal.hbs
+++ b/addon/components/au-modal.hbs
@@ -20,7 +20,7 @@
       tabindex="-1"
       {{focus-trap
         isActive=@modalOpen
-        additionalElements=(array '#ember-basic-dropdown-wormhole')
+        additionalElements=(array "#ember-basic-dropdown-wormhole")
         focusTrapOptions=(hash
           initialFocus=this.initialFocus
           allowOutsideClick=true
--- a/addon/components/au-modal.js
+++ b/addon/components/au-modal.js
@@ -2,6 +2,8 @@ import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 
+const FOCUS_TRAP_ADDITIONAL_ELEMENTS = ['#ember-basic-dropdown-wormhole'];
+
 export default class AuModal extends Component {
   constructor() {
     super(...arguments);
@@ -34,6 +36,12 @@ export default class AuModal extends Component {
 
   get initialFocus() {
     return this.args.initialFocus ?? '.au-c-modal__title';
+  }
+
+  get additionalElements() {
+    return FOCUS_TRAP_ADDITIONAL_ELEMENTS.filter(
+      (element) => document.querySelector(element) !== null
+    );
   }
 
   @action
--- a/app/styles/ember-appuniversum/_p-ember-power-select.scss
+++ b/app/styles/ember-appuniversum/_p-ember-power-select.scss
@@ -70,6 +70,9 @@ $ember-power-select-opened-border-radius: 0; // Border radius of the side of the
 $ember-power-select-search-input-border-radius: 0.3rem;
 $ember-power-select-multiple-option-border-radius: $ember-power-select-default-border-radius;
 
+// Z-index
+$ember-power-select-dropdown-z-index: 10000; // Z-index of the dropdown when it is not rendered in place (needs to be higher then z-index of the modal component)
+
 // Other
 $ember-power-select-focus-box-shadow: none;
 $ember-power-select-dropdown-margin: 0; // Margin between the dropdown and the trigger
@@ -388,6 +391,7 @@ $ember-power-select-multiple-option-line-height: 1.45;
   box-shadow: $ember-power-select-dropdown-box-shadow;
   overflow: hidden;
   color: $ember-power-select-text-color;
+  z-index: $ember-power-select-dropdown-z-index;
 }
 
 .ember-power-select-dropdown.ember-basic-dropdown-content--above {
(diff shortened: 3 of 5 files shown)


brenner-company:
@Windvis A few small questions:
- The following CSS change (https://github.com/appuniversum/ember-appuniversum/pull/391/files#diff-d9d5369a850518e2107794fe917092f0b863f734ad3d86e95677417024f7297aR391) should be needed to display the dropdown element of the `ember-focus-trap` above the `AuModal` instance. Do we provide this ourselves or is this something each destination project that want this, need to implement on their own?
- Do we adjust the Storybook docs regarding `AuModal` (overflow version), add a new one or leave these changes as a silent introduction?
- I don't think it is needed here to provide any additional test?

brenner-company:
I see a few tests regarding `AuModal` failing by the way. I'll give them a look!

addon/components/au-modal.hbs
```diff
@@ -20,6 +20,7 @@
       tabindex="-1"
       {{focus-trap
         isActive=@modalOpen
+        additionalElements=(array "#ember-basic-dropdown-wormhole")
```
brenner-company, line 23: Should this perhaps be an application variable/option?

brenner-company:
> I see a few tests regarding `AuModal` failing by the way. I'll give them a look!

I'm an idiot 🤦‍♂️ I forgot to implement an existence check on the actual elements list. Fixed.

addon/components/au-modal.js
```diff
@@ -2,6 +2,8 @@ import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 
+const FOCUS_TRAP_ADDITIONAL_ELEMENTS = ['#ember-basic-dropdown-wormhole'];
```
brenner-company, line 5: Should this perhaps be an application variable/option?

Windvis:
> @Windvis A few small questions:
> 
>     * The following CSS change (https://github.com/appuniversum/ember-appuniversum/pull/391/files#diff-d9d5369a850518e2107794fe917092f0b863f734ad3d86e95677417024f7297aR391) should be needed to display the dropdown element of the `ember-focus-trap` above the `AuModal` instance. Do we provide this ourselves or is this something each destination project that want this, need to implement on their own?

I think we already provide some power-select styles that the app can import so some extra styles makes sense?

>     * Do we adjust the Storybook docs regarding `AuModal` (overflow version), add a new one or leave these changes as a silent introduction?
I'm not sure. Sorry, bad memory. What was the issue again with the setup that's currently displayed in the overflow example? If the "new" setup makes more sense for most apps then it might be better to update the example?

>     * I don't think it is needed here to provide any additional test?
No I don't think so either.

addon/components/au-modal.js
Windvis: What did you have in mind exactly? I'm not sure what you mean.

addon/components/au-modal.js
```diff
@@ -36,6 +38,12 @@ export default class AuModal extends Component {
     return this.args.initialFocus ?? '.au-c-modal__title';
   }
 
+  get additionalElements() {
+    return FOCUS_TRAP_ADDITIONAL_ELEMENTS.filter(
+      (element) => document.querySelector(element) !== null
+    );
+  }
```
Windvis, line 45: I would just do that in the module scope since this never changes while the app is running but the getter can potentially run multiple times.

brenner-company: Not entirely sure what the module scope is. Is that within the `constructor`? Or even outside the component class?

brenner-company:
> I think we already provide some power-select styles that the app can import so some extra styles makes sense?

I think so 👍 I'll make the adjustments so you can give them a look.

> I'm not sure. Sorry, bad memory. What was the issue again with the setup that's currently displayed in the overflow example? If the "new" setup makes more sense for most apps then it might be better to update the example?

No problem. In the current situation, if you have a `ember-power-select` instance within a modal, you need to enable `@overflow` on that modal otherwise the dropdown of the `ember-power-select` instance can get clipped (as it is rendered in place with `@renderInPlace={{true}}` and not through a wormhole). But in some cases, you actually need `@overflow` disabled so you can scroll through the content of that modal. You can't mix those at the moment.

We could fix this within Kaleidos by disabling `@renderInPlace` on the `ember-power-select` instance, but then the `ember-focus-trap` functionality within the modal breaks. You can't focus the search (within the `ember-power-select` dropdown) as `ember-focus-trap` thinks: "That focus/tabbing is outside the modal, not doing that".

This change will fix that last problem. And with that all possible set-ups regarding modal and `ember-power-select` are possible (with `@overflow` not being necessary anymore). I guess we could remove the Storybook story for that & add a remark for all the developers using `ember-appuniversum` within the `#appuniversum` channel on Rocket chat.

> No I don't think so either.

✅

addon/components/au-modal.js
brenner-company: Nevermind that last idea.

brenner-company:
> > I think we already provide some power-select styles that the app can import so some extra styles makes sense?
> 
> I think so 👍 I'll make the adjustments so you can give them a look.

Done!

brenner-company:
@Windvis If you could give the remaining questions/remarks in here another look (if you have the time of course), that would be peachy. Thanks!

addon/components/au-modal.js
Windvis: Outside the component class, but nevermind, it doesn't matter that much 😄.

addon/components/au-modal.js
Windvis: Ideally apps could simply wrap the AuModal component and configure any app wide configs that way, but that isn't really possible yet with named blocks..

So a different way to configure this might be nice. The problem is that the configs we currently support are only temporary (until v3) and the same system doesn't work for v2 addons so I would rather not add extra options that way.

Looking up the environment config could work, but that's not really a good practice (anymore). I would have to think about how this can best be done but I think we can merge this as-is and add the option in a future PR.

Windvis:
@brenner-company Sorry I thought I answered all open question / remarks. I answered some more, can you let me know if I missed anything else? 😄

addon/components/au-modal.js
brenner-company: @Windvis No problem, it is always a good idea to accustom to good practices 😉

Like this?

CleanShot 2023-05-22 at 19 48 38@2x

Windvis: Yea, something like that, but I realized it might cause issues if the elements aren't there yet when the code is executed. It will also cause issues in apps that use FastBoot since `document` isn't a thing there. So keep it as-is! It was premature optimization anyways. If we notice it is slow we can cache the result of the getter.

PR #396, opened 2023-05-03
Disable autofill for AuDateInput components

Riadabd:
(Addresses DL-5081) Disables auto-fill on start and end date fields when creating a mandate which was causing issues with the validation code of the `Startdatum` field.


addon/components/au-date-input.hbs
```diff
@@ -7,8 +7,9 @@
...
     disabled={{@disabled}}
     placeholder="DD-MM-JJJJ"
+    autocomplete="off"
     ...attributes
     {{this.dateInput value=@value prefillYear=@prefillYear onChange=@onChange}}
   />
   <AuIcon @icon="calendar" @alignment="left" />
-</span>
\ No newline at end of file
+</span>
```
Windvis, line 15: I'm not 100% sure, but I think this diff means that a newline was added here? If so, that's not recommended for .hbs files (since that creates unneeded nodes in the AST / final template). Do you have an .editorconfig plugin setup? It should already handle that: https://github.com/appuniversum/ember-appuniversum/blob/cc01211b2d27c27f6ebd045da49113623badca61/.editorconfig#L16

Windvis: I think Prettier does the formatting (which follows the editorconfig file as well), so maybe the editorconfig plugin wasn't active yet on save? Maybe a restart is needed?

Windvis (approved):

addon/components/au-date-input.hbs
Riadabd: Prettier does remove the newline, but when I save again, VS Code re-inserts the newline at the end.

Issue #398, opened 2023-06-09
Incorrect flanders-sans font baseline on Windows environments


brenner-company:
When viewing the Storybook component/styleguide on a Windows environment, in my case Windows 10 with various browsers, the loaded `flanders-sans` font is displayed with a incorrect font baseline in certain situations (font-size, font-weight, etc).

The screenshot below is taken from the `Typography` page ([reference link](https://appuniversum.github.io/ember-appuniversum/?path=/story/utilities-typography--page)) on a Windows 10 machine with Chrome:

![2023-06-09 18_51_42-Utilities _ Typography - Page ⋅ Storybook](https://github.com/appuniversum/ember-appuniversum/assets/18174827/bde1cdc3-65e0-4f83-a90c-6bfdd629cfd0)

FYI: We encountered this issue before within Kaleidos (in more detail, our own CSS package `au-kaleidos-css`) and found a fix. When we replaced the font files with the ones used on https://www.vlaanderen.be/, the issue disappeared. The exact cause of this particular issue was not found (must be linked most likely to the font generator process).

Is there a way to replace the existing font files on the cloudfront cdn that `ember-appuniversum` uses or upload a different (correct) set?

Windvis:
I know nothing about the CDN setup, sorry. Maybe we can bundle the fonts with Appuniversum?

brenner-company:
> I know nothing about the CDN setup, sorry. Maybe we can bundle the fonts with Appuniversum?

I think that is the only possible way. I made a draft PR for you to give it look at https://github.com/appuniversum/ember-appuniversum/pull/399. Thanks!

PR #399, opened 2023-06-22
Addition of bundled (flanders-serif & flanders-sans) font files

brenner-company:
In regard to https://github.com/appuniversum/ember-appuniversum/issues/398: bundling of updated/corrected `flanders-serif` & `flanders-sans` font files with `ember-appuniversum`.


brenner-company:
@Windvis Currently, I'm not aware of any method to automatically load in the font files within a destination project like Kaleidos. I assume you have to specifically import the files within `ember-cli-build.js` with `app.import` of `Funnel`?

Windvis:
Since the fonts are stored in the public folder, Ember should automatically put them under the `dist/addon-name/*` path in the dist folder.

So for Appuniversum it would be at `dist/@appuniversum/ember-appuniversum/fonts/*`. So I _think_ we can use a config like this: `$au-font-location: "../@appuniversum/ember-appuniversum/fonts";`

This seems to work when I quickly tested it in Loket.

brenner-company:
> Since the fonts are stored in the public folder, Ember should automatically put them under the `addon-name/*` path in the dist folder.
> 
> So for Appuniversum it would be at `dist/@appuniversum/ember-appuniversum/fonts/*`. So I _think_ we can use a config like this: `$au-font-location: "../@appuniversum/ember-appuniversum/fonts";`
> 
> This seems to work when I quickly tested it in Loket.

Also tested this within Kaleidos and this works perfectly. Didn't know you could do that 👍

brenner-company:
I noticed that Chromatic is triggering a lot of diff alerts: most of them were caused by the minute difference in that way that font files were rendered, but there is also a spacing/dimensions issue.

CleanShot 2023-07-05 at 13 36 48@2x

I guess these elements may be using character specific spacing/dimensions? I'll give it a look!

Windvis (approved):

brenner-company:
> I noticed that Chromatic is triggering a lot of diff alerts: most of them were caused by the minute difference in that way that font files were rendered, but there is also a spacing/dimensions issue.
> 
> I guess these elements may be using character specific spacing/dimensions? I'll give it a look!

I guess this is just a fluke within Chromatic because when I compare the online version of Storybook with my local version, I don't see the same visual issue. Just the minute difference in line height.

CleanShot 2023-07-06 at 12 40 52@2x
CleanShot 2023-07-06 at 12 40 30@2x

brenner-company:
Nevermind, I just spotted there is a difference within Firefox. I'll check it out 👀

brenner-company:
> Nevermind, I just spotted there is a difference within Firefox. I'll check it out 👀

After a bit of research, the change in width of the input is purely caused by the (change in) font family used for the input field, so I'll just accept all the visual changes within Chromium 👍

Windvis:
Sounds good. Thanks for looking into it!

Issue #397, opened 2023-06-09
Addition of action button within toaster alert


brenner-company:
Within Kaleidos we would love to replace our current & legacy toaster (service) with the one provided within `ember-appuniversum`, but at the moment we're missing the option to add an action button to the bottom of the toaster alert (see quick mockup below).

<img width="324" alt="CleanShot 2023-06-09 at 18 07 48@2x" src="https://github.com/appuniversum/ember-appuniversum/assets/18174827/307a694d-0a5e-4855-90cc-b31c61768a37">

Is this something that can be added to the current toaster component? If yes, can I help with that?

brenner-company:
> I guess we could go for flexibility and allow users to pass in the alert component that needs to be rendered, similar to ember-promise-modals? https://github.com/mainmatter/ember-promise-modals

@Windvis I'm not entirely sure how this should be done/structured? Can you perhaps reference a example within `ember-promise-modals`? Much Thanks!

Windvis:
In ember-promise-modals you have to tell the addon which component it should use, for example: https://github.com/mainmatter/ember-promise-modals#embroider-optimized-apps

`this.modals.open(ConfirmationModal);`

We could do something similar to give users complete control over the rendered toast. Some pseudo-example:

`this.toaster.show(SomeToastComponentWithAButton, options);`

Anyway, I'm not saying we have to go this route. It depends on how flexible we want this to be I guess.

I'm not sure what you had in mind yourself but adding a button with an icon requires quite a bit of extra options if you want everything configurable.

```js
this.toaster.notify('message', 'Notify', {
  // other toast options
  // ..
  actionIcon: 'reload', 
  actionSkin: 'link', 
  actionText: 'Lorem ipsum', 
  actionHandler: someFunction,
  // possibly more options
});
```

In the custom component version most of these options would be part of the component itself, so less things need to be passed in through the options object. And we wouldn't have to make changes here if something extra is needed in the component.

PR #401, opened 2023-07-10
Addition of action button within toaster (& alert) component

brenner-company:
Addition of the following functionality for the `AuToaster` & `AuAlert` components:
* `AuAlert`: inclusion of a possible action block. I've added this because we'll need this anyway in the near future within Kaleidos & this makes implementing functionality within `AuToaster` a bit easier. You can check Storybook for a new example regarding that.
* `AuToaster`: inclusion of possible action configuration within the toaster options.

What is currently missing (within draft):
* Some finalization & code clean-up (see remarks)
* Accompanying tests

PS: @Windvis, I've checked your comment at https://github.com/appuniversum/ember-appuniversum/issues/397#issuecomment-1621658212. That seems like a great method to implement similar changes, but I don't think this has much advantage over the current version/changes within this draft? Feel free to correct me if this is not the case!

### DIFF
--- a/addon/components/au-toaster.hbs
+++ b/addon/components/au-toaster.hbs
@@ -1,17 +1,45 @@
 {{#if this.toaster.toasts.length}}
   <div class="au-c-toaster {{this.position}}">
     {{#each this.toaster.toasts as |toast|}}
-      <AuAlert
-        @title={{toast.title}}
-        @skin={{toast.options.type}}
-        @icon={{toast.options.icon}}
-        @size="small"
-        @closable={{toast.options.closable}}
-      >
-        {{#if toast.message}}
-          <p>{{toast.message}}</p>
-        {{/if}}
-      </AuAlert>
+      {{#if toast.options.action}}
+        <AuAlert
+          @title={{toast.title}}
+          @skin={{toast.options.type}}
+          @icon={{toast.options.icon}}
+          @size="small"
+          @closable={{toast.options.closable}}
+          @isAToast={{true}}
+        >
+          <:default>
+            {{#if toast.message}}
+              <p>{{toast.message}}</p>
+            {{/if}}
+          </:default>
+          <:action>
+            <AuButton
+              @skin="link"
+              @icon={{toast.options.action.icon}}
+              @alert={{toast.options.action.alert}}
+              {{on "click" toast.options.action.handler}}
+            >
+              {{toast.options.action.text}}
+            </AuButton>
+          </:action>
+        </AuAlert>
+      {{else}}
+        <AuAlert
+          @title={{toast.title}}
+          @skin={{toast.options.type}}
+          @icon={{toast.options.icon}}
+          @size="small"
+          @closable={{toast.options.closable}}
+          @isAToast={{true}}
+        >
+          {{#if toast.message}}
+            <p>{{toast.message}}</p>
+          {{/if}}
+        </AuAlert>
+      {{/if}}
     {{/each}}
   </div>
 {{/if}}
--- a/addon/components/au-toaster.hbs
+++ b/addon/components/au-toaster.hbs
@@ -8,7 +8,7 @@
           @icon={{toast.options.icon}}
           @size="small"
           @closable={{toast.options.closable}}
-          @isAToast={{true}}
+          class="au-c-alert--toast"
         >
           <:default>
             {{#if toast.message}}
@@ -33,7 +33,7 @@
           @icon={{toast.options.icon}}
           @size="small"
           @closable={{toast.options.closable}}
-          @isAToast={{true}}
+          class="au-c-alert--toast"
         >
           {{#if toast.message}}
             <p>{{toast.message}}</p>
--- a/addon/components/au-alert.hbs
+++ b/addon/components/au-alert.hbs
@@ -1,9 +1,6 @@
 {{#if this.isVisible}}
   <div
-    class="au-c-alert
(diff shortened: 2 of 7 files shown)


addon/components/au-toaster.hbs
```diff
@@ -1,17 +1,45 @@
...
-        {{/if}}
-      </AuAlert>
+      {{#if toast.options.action}}
+        <AuAlert
+          @title={{toast.title}}
+          @skin={{toast.options.type}}
+          @icon={{toast.options.icon}}
+          @size="small"
+          @closable={{toast.options.closable}}
+          @isAToast={{true}}
```
brenner-company, line 11: I'm not entirely sure this is the best (argument) name to differentiate the layout type within `AuAlert`?

Windvis: Can't we just pass a css class to the component?

addon/components/au-toaster.hbs
```diff
@@ -1,17 +1,45 @@
...
+        >
+          <:default>
+            {{#if toast.message}}
+              <p>{{toast.message}}</p>
+            {{/if}}
+          </:default>
+          <:action>
+            <AuButton
+              @skin="link"
```
brenner-company, line 20: FYI: I've locked the skin here to `link` as I don't think the other skins match the toaster context or would steal to much visual focus.

addon/components/au-toaster.hbs
brenner-company: Oh, that is a better/simpler solution then mine 👍 I'll change it.

Windvis:
> PS: @Windvis, I've checked your comment at https://github.com/appuniversum/ember-appuniversum/issues/397#issuecomment-1621658212. That seems like a great method to implement similar changes, but I don't think this has much advantage over the current version/changes within this draft? Feel free to correct me if this is not the case!

The main reason I preferred the flexible version is that every project can then do what they want. I'm not sure that 1 action button covers all the use cases. It probably does, but when it doesn't we will have to do some weird API changes to make it possible. With the component version we still had that as a fallback.

I can see why a stricter API is a good thing though. All projects would do it in a similar way and it's easier to use.


I looked at the Webuniversum version again and it seems they do actions slightly different there: https://overheid.vlaanderen.be/webuniversum/v3/documentation/js-components/vl-ui-toaster (although they do only show 1 action button).

Maybe we should try to align the design?

addon/components/au-toaster.hbs
```diff
@@ -1,17 +1,45 @@
...
+              @icon={{toast.options.action.icon}}
+              @alert={{toast.options.action.alert}}
+              {{on "click" toast.options.action.handler}}
```
Windvis, line 23: Should it be possible to show a loading state as well? If so, that might complicate things.

brenner-company:
> The main reason I preferred the flexible version is that every project can then do what they want. I'm not sure that 1 action button covers all the use cases. It probably does, but when it doesn't we will have to do some weird API changes to make it possible. With the component version we still had that as a fallback.
> 
> I can see why a stricter API is a good thing though. All projects would do it in a similar way and it's easier to use.
>
I can give this a look tomorrow and create a version within a new branch. But I'm not sure how much time I can spend on this because this effort is getting more and more in the `ember-appuniversum` (as this is become a bit of a refactor?) and not Kaleidos territory. I'll keep you posted.

Functionality wise, if I understand you correctly I can rework `addon/services/toaster.js` a bit so we still have the `.notify()`, `.success()`, etc. functions but we a new `.show()` one? Or do we remove those and use `.show()` altogether?

And do we add the new custom component (toast with action button) to `ember-appuniversum` itself? Together with the alignment with the design of course.
>
> I looked at the Webuniversum version again and it seems they do actions slightly different there: https://overheid.vlaanderen.be/webuniversum/v3/documentation/js-components/vl-ui-toaster (although they do only show 1 action button).
> 
> Maybe we should try to align the design?
That is no a bad idea I guess? So, if the toaster needs to be the warning version/skin with an action button, that would be an exception to the rule and would be the perfect case for a custom toaster component within the destination project?

Windvis:
> I can give this a look tomorrow and create a version within a new branch.

No, no need, keep it like this. We can always add the component version in the future and deprecate the action options if needed.

>  But I'm not sure how much time I can spend on this because this effort is getting more and more in the ember-appuniversum (as this is becoming a bit of a refactor?) and not Kaleidos territory. I'll keep you posted.

Well, this is the hard part for me. My personal opinion is that Appuniversum should be as generic as possible and features should only be added if it will be widely used. 

To me it was not clear if this "action in a notification" is a generic feature or something only Kaleidos needed. So my preference was a generic / lower level solution which apps could build on, without having to ship the opinionated version as part of Appuniversum itself. But it seems there are examples of this case in Webuniversum, so that seems good enough reason to ship the higher level implementation instead.

Windvis:
(let me know if this is blocked by me in some way)

brenner-company:
I think I'll need to extend this PR a tad more with additional functionality regarding the button skin, but I'll pick that up when I get back.

Thanks for the follow-up!

brenner-company:
I've just heard that we'll also need a `AuLinkExternal` inside another variant of the toast, so I'm afraid the current version of the PR won't be sufficient anymore.

Is ok for you if I give the component version (mentioned here: https://github.com/appuniversum/ember-appuniversum/issues/397#issuecomment-1621658212) a look tomorrow?

PR #412, opened 2023-07-27
Removal of backdrop click close trigger & addition of closable boolean for modal component

### LINKED ISSUE #411: Addition of adjustable close triggers & closable boolean for modal component
Within Kaleidos we could use the following functionality for the `AuModal` component:
* The addition of an adjustable close trigger list, similar to what is being used here: https://buefy.org/documentation/modal/#api-view (see `can-cancel`). `escape`, `x` & `outside` should suffice as possible options.
* The addition of a `closable` boolean so that the closing or canceling of the modal can be blocked.

I've made a draft PR as a proof-of-concept that can be checked out if needed.

brenner-company:
An enhancement for the `AuModal` component regarding the following: 
* ~~The addition of an adjustable close trigger list.~~
* The removal of the escape key close trigger (as mentioned here https://github.com/appuniversum/ember-appuniversum/issues/411#issuecomment-1659786997)
* The addition of a `closable` boolean.

More information can be found inside the complementary FR: https://github.com/appuniversum/ember-appuniversum/issues/411.

Closes #411


addon/components/au-modal.hbs
```diff
@@ -43,15 +43,18 @@
...
-        <button
-          class="au-c-modal__close"
-          type="button"
-          data-test-modal-close
-          {{on "click" this.closeModal}}
-        >
-          <AuIcon @icon="cross" @size="large" />
-          <span class="au-u-hidden-visually">Venster sluiten</span>
-        </button>
+        {{#if this.closeTriggers.x}}
```
brenner-company, line 46: I'm not entirely sure if making it possible to remove the close button (as a trigger) in the modal header is good practice accessibility/UX wise?

addon/components/au-modal.js
```diff
@@ -16,6 +21,8 @@ export default class AuModal extends Component {
       'au-modal: No target element was found. Please add the `<AuModalContainer />` component where appropriate.',
       this.destinationElement
     );
+
+    this.closeTriggers = this.args.closeTriggers ? {...CLOSE_TRIGGERS, ...this.args.closeTriggers} : CLOSE_TRIGGERS;
```
brenner-company, line 25: I've used `closeTriggers` as a variable name here, but I'm not entirely sure this is the best possible option. Something like `canClose` could also suffice?

brenner-company:
Note: I'll add some test covering the new functionality if this PR gets more tangible or gets finalized.

brenner-company:
@Windvis Should I also add a test covering `@closable` (when using close button within modal header & escape key)?

brenner-company:
@Windvis: I've added the test (https://github.com/appuniversum/ember-appuniversum/pull/412/commits/7c90f022896abba45c710f772ac60a60cfa4a1f8) in the meantime, but with two remarks/questions:
* The test fails at the moment because of the following: when settings `@closable` to `false` and providing a blank modal with no title, body & footer, there is no element for focus-trap to focus to so it throws an error (as expected). Two things I noticed or want to get cleared up a bit:
  * When looking at the getter `initialFocus` I see that `.au-c-modal__title` is returned as a default, but the element is actually not focusable (to focus-trap). Should this be corrected? Tying in with the failing test above, this could be a possible solution (by just adding a title to the test modal).
  * But that is still not completely covering blank modals (even without a title). Is it maybe an idea to provide a `fallbackFocus` element (as mentioned here: https://github.com/focus-trap/focus-trap#createoptions) to make sure focus-trap doesn't throw an error? Maybe the modal element itself (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations mentions it as a possibility, though not a common one)? 
* Should I also add the escape key press to this test or is the clicking of the close button (within the header) sufficient for this?

@Dietr If you want to chime in on the first part of the two questions, feel free!

brenner-company:
@Windvis Is possible to give this a (finalizing) look in the following days? (I'm asking because this missing feature is blocking a story within Kaleidos)

addon/components/au-modal.hbs
```diff
@@ -44,10 +38,11 @@
           {{/if}}
         </h1>
         <button
-          class="au-c-modal__close"
+          class="au-c-modal__close {{unless this.isClosable 'is-disabled'}}"
```
Windvis, line 41: UX question: is disabling the button better than not showing it at all? Does it depend on the use-case?

Also, while testing it was hard to see the difference between the disabled and non-disabled states (other than the cursor change).

Windvis:
> * The test fails at the moment because of the following: when settings `@closable` to `false` and providing a blank modal with no title, body & footer, there is no element for focus-trap to focus to so it throws an error (as expected). Two things I noticed or want to get cleared up a bit:
>   
>   * When looking at the getter `initialFocus` I see that `.au-c-modal__title` is returned as a default, but the element is actually not focusable (to focus-trap). Should this be corrected? Tying in with the failing test above, this could be a possible solution (by just adding a title to the test modal).

I don't know, sorry. I have little experience with element focusing. Why don't the other tests fail? focus trap tries to focus the close button or something? It does sound strange to set an element that can't be focused as the default, not sure about the history of the implementation.

>   * But that is still not completely covering blank modals (even without a title). Is it maybe an idea to provide a `fallbackFocus` element (as mentioned here: https://github.com/focus-trap/focus-trap#createoptions) to make sure focus-trap doesn't throw an error? Maybe the modal element itself (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations mentions it as a possibility, though not a common one)?

You want to prevent that focus-trap throws exceptions because it's an implementation detail of the component? So if I understand it correctly, the difference between initialFocus and fallbackFocus is that the latter won't throw an error if there are no focusable elements inside? Maybe we should use that instead of the initialFocus default value then?


> * Should I also add the escape key press to this test or is the clicking of the close button (within the header) sufficient for this?

I think it might be nice to test the escape key press as well, but that can be copy pasted from the other test.

Windvis:
Btw, the title / description is a bit confusing. It's the backdrop close that was removed, right? The others are just conditional based on `@closable`?

addon/components/au-modal.hbs
brenner-company: > UX question: is disabling the button better than not showing it at all? Does it depend on the use-case?

I think making an interaction component invisible when it or its parent is disabled will be confusing for the user.

> Also, while testing it was hard to see the difference between the disabled and non-disabled states (other than the cursor change).

Yup, I noticed that too. I think I'll give it a go today with a lighter `disabled` color value.

brenner-company:
> Btw, the title / description is a bit confusing. It's the backdrop close that was removed, right? The others are just conditional based on `@closable`?

Damn, you're correct, I'll correct that!

tests/integration/components/au-modal-test.js
```diff
@@ -202,12 +201,13 @@ module('Integration | Component | au-modal', function (hooks) {
...
     await render(hbs`
       <AuModalContainer />
-      <AuModal @modalOpen={{this.isOpen}} @closeModal={{this.handleClose}}></AuModal>
+      <AuModal @modalOpen={{this.isOpen}} @closeModal={{this.handleClose}} @closable={{false}}></AuModal>
     `);
 
-    let backdrop = document.querySelector(MODAL.BACKDROP);
-    await click(backdrop);
-    assert.strictEqual(timesCalled, 1);
+    assert.dom(MODAL.CLOSE).isDisabled();
```
brenner-company, line 207: I've had to change this from an actual `click()` to just a `isDisabled()` check because disabled elements are apparently not clickable within tests. Cypress has something along `click({force: true})` but I haven't found anything similar yet.

brenner-company: If you can think of a better or more proper solution, please tell!

addon/components/au-modal.hbs
```diff
@@ -18,6 +18,7 @@
         additionalElements=this.additionalElements
         focusTrapOptions=(hash
           initialFocus=this.initialFocus
+          fallbackFocus=this.fallbackFocus
```
brenner-company, line 21: Weird thing I noticed here is that defining `fallbackFocus` also fixes the ability to focus the `h1` title element with focus-trap if it is empty (as mentioned here: https://github.com/appuniversum/ember-appuniversum/pull/412#issuecomment-1674523425). Not sure why 🤷‍♂️

brenner-company:
I've made a small recording of the 'new' focus behaviour within the modal (with added dotted outlines for modal title & modal for easier focus visualisation):

https://github.com/appuniversum/ember-appuniversum/assets/18174827/d45ba06e-f741-4af2-a390-b86c314bb0b3

Seems good right? (@Dietr, feel free to interject if needed)

Windvis (approved):

Issue #413, opened 2023-07-28
Addition of clear action within toaster service


brenner-company:
In addition to https://github.com/appuniversum/ember-appuniversum/issues/397: 

To replace our current & legacy toaster (service) with the one provided within `ember-appuniversum` we would also need a clear function within the toaster service. This would be used in specific situations where the toast is not applicable anymore after being displayed.

I've made a draft PR as a proof-of-concept that can be checked out if needed.

PR #414, opened 2023-07-28
Addition of clear action within toaster service

brenner-company:
The addition of a clear function within the `toaster` service.

More information can be found inside the complementary FR: https://github.com/appuniversum/ember-appuniversum/issues/413.

### DIFF
--- a/addon/services/toaster.js
+++ b/addon/services/toaster.js
@@ -21,9 +21,7 @@ export default class ToasterService extends Service {
     if (toast.options.timeOut) {
       yield timeout(toast.options.timeOut);
 
-      if (this.toasts.includes(toast)) {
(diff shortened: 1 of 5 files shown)


Windvis:
Sounds good 👍. Maybe also add the code snippet to the Storybook? I don't think we documented that the display methods return "toast" objects yet.

brenner-company:
Sounds good! I'll finalize this PR when I get back.

brenner-company:
> I added some tweaks in #418 that targets this PR. 

Thanks for the help!

> Should be good to go I think (unless you have strong opinions about the method rename 😄).

Nono, I just used `clear` because it was the name we used in Kaleidos to clear the toasts. Sounds good!

PR #418, opened 2023-08-14
Some toaster tweaks

Windvis:
- rename the method to `close` which seems like a better fit
- add some documentation
- rework the existing documentation


brenner-company (approved):

brenner-company:
> * rename the method to `close` which seems like a better fit

That is, indeed, a better match 👍 Thanks for providing the documentation!

PR #420, opened 2023-08-15
Add support for custom toaster components (@brenner-company version)

brenner-company:
See https://github.com/appuniversum/ember-appuniversum/pull/419 for more information.

### DIFF
--- a/addon/components/au-toast.hbs
+++ b/addon/components/au-toast.hbs
@@ -1,0 +1,11 @@
+<AuAlert
+  @title={{@title}}
+  @skin={{@options.type}}
+  @icon={{@options.icon}}
+  @size="small"
+  @closable={{@options.closable}}
+>
+  {{#if @message}}
+    <p>{{@message}}</p>
+  {{/if}}
+</AuAlert>
--- a/addon/components/au-toaster.hbs
+++ b/addon/components/au-toaster.hbs
@@ -1,17 +1,12 @@
 {{#if this.toaster.toasts.length}}
   <div class="au-c-toaster {{this.position}}">
     {{#each this.toaster.toasts as |toast|}}
-      <AuAlert
-        @title={{toast.title}}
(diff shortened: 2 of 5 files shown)


addon/components/au-toast.hbs
Windvis, line 1: Do you think this component is needed? It seems to add little value over the `AuAlert`?

addon/components/au-toaster.hbs
```diff
@@ -1,17 +1,12 @@
...
-        @icon={{toast.options.icon}}
-        @size="small"
-        @closable={{toast.options.closable}}
-      >
-        {{#if toast.message}}
-          <p>{{toast.message}}</p>
-        {{/if}}
-      </AuAlert>
+      {{#let
+        (component (ensure-safe-component toast.name) title=toast.title message=toast.message options=toast.options)
```
Windvis, line 5: As mentioned in the other PR, `component` and `ensure-safe-component` shouldn't be needed here.

I went with passing only the options (and a close action) into the component since the `show` method doesn't accept a message and title and users have to pass everything through options anyway.

addon/components/au-toast.hbs
brenner-company: See https://github.com/appuniversum/ember-appuniversum/pull/419#issuecomment-1680367652.

addon/components/au-toaster.hbs
brenner-company: Also see https://github.com/appuniversum/ember-appuniversum/pull/419#issuecomment-1680367652.

> I went with passing only the options (and a close action) into the component since the `show` method doesn't accept a message and title and users have to pass everything through options anyway.

Seems a tad more logical that way 👍 Glad to be learning!

brenner-company:
I'll just close this one down as all work is being continued inside https://github.com/appuniversum/ember-appuniversum/pull/419.

PR #421, opened 2023-08-24
Update dependencies

Windvis:
This updates or widens the range for direct dependencies. Projects can then install the newer version and Appuniversum will follow it, allowing deduplication.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@storybook/addon-storysource": "^6.5.7",
     "@storybook/addons": "^6.5.7",
     "@storybook/builder-webpack5": "^6.5.7",
-    "@storybook/ember": "^6.5.7",
+    "@storybook/ember": "6.5.14",
     "@storybook/ember-cli-storybook": "^0.6.0",
     "@storybook/manager-webpack5": "^6.5.7",
     "@storybook/theming": "^6.5.7",
--- a/package.json
+++ b/package.json
@@ -56,19 +56,19 @@
     "@floating-ui/dom": "^1.1.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -84,7 +88,7 @@
     "@storybook/addon-storysource": "^6.5.7",
     "@storybook/addons": "^6.5.7",
     "@storybook/builder-webpack5": "^6.5.7",
-    "@storybook/ember": "^6.5.7",
+    "@storybook/ember": "6.5.14",
```
Windvis, line 91: Storybook fails to build when we install a newer version, I haven't looked into it yet.

package.json
```diff
@@ -56,26 +56,30 @@
     "@floating-ui/dom": "^1.1.0",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
-    "@zestia/ember-auto-focus": "^4.2.0",
+    "@zestia/ember-auto-focus": "^4.2.0 || ^5.0.0",
```
Windvis, line 59: I'm not sure if this is breaking or not 🤔. My intention was that apps that _do_ want v5, can install it, and Appuniversum would follow (if deduped), but not every app will have this installed. I'm not how the package managers will behave either when they see this range. They could in theory install the latest version.

package.json
```diff
@@ -132,7 +136,7 @@
     "ember-power-select": "2.x || 3.x || 4.x || 5.x || 6.x || 7.x"
   },
   "peerDependencies": {
-    "ember-source": "^3.28.0 || ^4.0.0"
+    "ember-source": "^3.28.0 || ^4.0.0 ||^5.0.0"
```
Windvis, line 139:
```suggestion
    "ember-source": "^3.28.0 || ^4.0.0 || ^5.0.0"
```

PR #428, opened 2023-09-13
Add a `{{au-inputmask}}` modifier

Windvis:
This is a simple wrapper around the inputmask library. This can be used to replace the masking functionality in the `AuInput` component, so we can now deprecate it and make the component simpler.

Separating it has the advantage that it won't be shipped under Embroider if it's not used anywhere.

This doesn't include docs yet since the AuInput component needs some changes first.

## Usage
```hbs
<input {{au-inputmask options=(hash mask="9.9.9")}} {{on "change" this.handleChange}} />
```

```js
export default class SomeComponent extends Component {
  handleChange = (event) => {
    console.log(event.target.inputmask.unmaskedvalue());
  }
}
```

`options` are simply passed into the Inputmask instance, which are documented here: https://robinherbots.github.io/Inputmask/#/documentation#options

Inputmask attaches itself to the element on the `inputmask` property, so any regular event handlers can be used to read out the data.

### DIFF
--- a/addon/modifiers/au-inputmask.js
+++ b/addon/modifiers/au-inputmask.js
@@ -1,0 +1,15 @@
+import { modifier } from 'ember-modifier';
+import Inputmask from 'inputmask';
+
+export default modifier(
+  function auInputmask(element, positional, { options = {} }) {
+    let inputmask = new Inputmask(options);
+
+    inputmask.mask(element);
+
+    return () => {
+      element.inputmask?.remove();
+    };
+  },
+  { eager: false },
+);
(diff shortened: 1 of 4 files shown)


addon/modifiers/au-inputmask.js
```diff
@@ -0,0 +1,17 @@
+import { modifier } from 'ember-modifier';
+import Inputmask from 'inputmask';
+
+export default modifier(
+  function auInputmask(element, positional, { options }) {
```
Windvis, line 5: should default to {} so we don't throw an error.

PR #429, opened 2023-09-13
Deprecate some `AuInput` features

### LINKED ISSUE #250: Deprecate the `Input` component usage in `AuInput`

Windvis:
This PR deprecates the 2-way-binding and input masking functionality in the `AuInput` component. This makes the component more flexible and easier to maintain.

It also means the inputmask library won't be bundled (when using Embroider "optimized") if the project doesn't use the masking feature (and doesn't use the `AuDateInput` component since that also uses it internally) reducing the bundle size by quite a bit.

Closes #250 

### 2-way-binding

To resolve the deprecation warnings, simply use the attribute versions and the `{{on}}` modifier instead.

```diff
- <AuInput @value={{this.someValue}} @type="number" />
+ <AuInput value={{this.someValue}} type="number" {{on "change" this.handleChange}} />
```

### input masking
The masking functionality has now moved to the [`{{au-inputmask}}` modifier](https://github.com/appuniversum/ember-appuniversum/pull/428). To resolve the deprecations, remove all masking related arguments, apply the modifier and pass them in there instead.

```diff
- <AuInput @mask="9.9.9" @maskPlaceholder="_" @onChange={{this.handleChange}} />
+ <AuInput {{au-inputmask options=(hash mask="9.9.9" placeholder="_")}} {{on "change" this.handleChange}} />
```

**Important:** The `@onChange` action returns the unmasked value. In order to match that behavior with the `{{on}}` modifier, the handler needs to be adjusted.

```diff
- @action onChange(newValue) { this.value = newValue }
+ @action onChange(event) { this.value = event.target.inputmask.unmaskedvalue(); }
```


addon/components/au-input.js
```diff
@@ -12,6 +12,23 @@ export default class AuInput extends Component {
       !this.args.onChange ||
         (typeof this.args.onChange === 'function' && this.isMasked),
     );
+
+    deprecate(
+      '[AuInput] The `@value` argument is deprecated. Use the `value` attribute and the `{{on}}` modifier instead.',
+      this.args.value === undefined,
```
Windvis, line 18: This won't fail if the consumer provides `undefined` as a value, we should use `'value' in this.args` instead

PR #440, opened 2023-10-05
.gjs conversion

Windvis:
This sets up ember-template-imports and converts most components to .gjs files.

Github doesn't have syntax highlighing support for .gjs yet since it requires a certain amount of files in (public?) repos. With this change we're helping a bit 😄.

This is an internal change, consumers shouldn't notice anything different and if they do it's most likely a bug.


Windvis:
The linting faillure is because of node 14.. We should bump it to 18 for the dev / CI setup and fix issues for node 14 when we run into them (until the next major, which should be released in the near future).

Windvis:
It seems there is an issue with accidental leading and trailing whitespace due to the conversion. I'm trying to figure out what the best way to deal with this is.

PR #449, opened 2023-10-16
Add vote star filled icon

lagartoverde:
Added this icon provided by Saska as we need it on the CLBV app, would be great if this can be released quickly :)


Windvis:
It seems we also keep a list in the dummy app: https://github.com/appuniversum/ember-appuniversum/blob/master/tests/dummy/app/components/au-icon-list.js

We should probably refactor that, but could you add it in there as well for now?

Windvis:
Also, doesn't this imply that there should also be an "empty" vote star icon? 😄

lagartoverde:
> Also, doesn't this imply that there should also be an "empty" vote star icon? 😄

Well, that is left as future work :P

Windvis:
> Well, that is left as future work :P

If that icon is already designed I would add it to this PR as well, just so we don't have to create multiple releases for icons 😄.

PR #454, opened 2023-12-07
Update dependencies to newest majors matching our support range

Windvis:
This bumps dependencies to the newest major versions supported by our support range.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "ember-data-table": "^2.1.0",
     "ember-file-upload": "^7.0.3",
     "ember-focus-trap": "^1.1.0",
-    "ember-modifier": "^3.2.7",
+    "ember-modifier": "^4.1.0",
     "ember-template-imports": "^3.4.2",
     "ember-test-selectors": "^6.0.0",
     "ember-truth-helpers": "^3.1.1 || ^4.0.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -65,24 +65,23 @@
...
     "ember-data-table": "^2.1.0",
-    "ember-file-upload": "^7.0.3",
+    "ember-file-upload": "^8.4.0",
     "ember-focus-trap": "^1.1.0",
-    "ember-modifier": "^3.2.7",
+    "ember-modifier": "^4.1.0",
     "ember-template-imports": "^3.4.2",
     "ember-test-selectors": "^6.0.0",
-    "ember-truth-helpers": "^3.1.1",
+    "ember-truth-helpers": "^3.1.1 || ^4.0.3",
```
Windvis, line 79: v3 is still needed because ember-data-table doesn't support v4 yet.

package.json
```diff
@@ -140,13 +139,15 @@
...
+    "tracked-built-ins": "^3.3.0",
     "webpack": "^5.88.2"
   },
   "optionalDependencies": {
     "ember-power-select": "2.x || 3.x || 4.x || 5.x || 6.x || 7.x"
   },
   "peerDependencies": {
-    "ember-source": "^4.8.0 || ^5.0.0"
+    "ember-source": "^4.8.0 || ^5.0.0",
+    "tracked-built-ins": "^3.3.0"
```
Windvis, line 150: ember-file-upload v8 lists it as a peerDependency so we do the same. New Ember projects will have this as part of the blueprint.

PR #460, opened 2024-01-16
Set up Glint and TypeScript for development

Windvis:
Add Glint and TypeScript (.gts and .ts) support to the addon, so we can start converting all the code. When everything is converted we can then publish the types.


Windvis:
I'm going to convert this to a "support gts / ts as a development tool" PR. Converting everything at once takes a long time, so it would be better to split that off and do it incrementally when we have time. Once everything is converted we can make the types public for consumers.

package.json
```diff
@@ -28,6 +44,9 @@
     "lint:hbs:fix": "ember-template-lint . --fix",
     "lint:js": "eslint . --cache",
     "lint:js:fix": "eslint . --fix",
+    "lint:types": "glint",
+    "prepack": "glint -d",
```
Windvis, line 48: This means we technically will already be publishing types once we add a .ts or .gts file, but I don't think that's a huge issue unless we document how to use them in apps. We consider these private until the docs are updated.

This makes testing the types easier and adventurous consumers can start using them (with the knowledge that things might break between releases).

tsconfig.declarations.json
Windvis, line 1: We can probably delete this file since it's not used, but it's part of the blueprint, and maybe Glint will support separate files in the future.

PR #464, opened 2024-02-02
`AuLoader` updates

### LINKED ISSUE #456: `AuLoader` visible message version
Webuniversum has the [vl-ui-loader component](https://overheid.vlaanderen.be/webuniversum/v3/documentation/atoms/vl-ui-loader) which is similar to Appuniversum's `AuLoader` but also allows for a visible loading message.

More projects are starting to need that functionality so we should provide it as part of the AuLoader component.

In Loket we've implemented a [custom component](https://github.com/lblod/frontend-loket/blob/development/app/components/page-loader.hbs) that provides the basic loader setup from vl-ui-loader.

**Open questions:**
1. Do we want the inline version?
2. Do we want the dark version?

Windvis:
This updates the `AuLoader` component so it matches the Webuniversum [`<vl-ui-loader>` component](https://overheid.vlaanderen.be/webuniversum/v3/documentation/atoms/vl-ui-loader/).

Notable changes:
- visual loading message by default
- an option to hide the loading message, if needed
- inline loading message variant
- the "old" version is now deprecated

Closes #456 

## Migration guide
The existing options that aren't supported by the `<vl-ui-loader>` component are now deprecated.

### No arguments / default loading message

If you were using the `AuLoader` component without any arguments the component would use a (visually hidden) default message. You can achieve a similar result by passing the previous default message as the block content.

```hbs
{{! Before }}
<AuLoader />

{{! After }}
<AuLoader @hideMessage={{true}}>Aan het laden</AuLoader>
```

### `@message`
Use the block content of the component to pass the loading message. If you want this loading message to only be visible for screenreaders you can combine this with the `@hideMessage` argument.

**Before:**
```hbs
{{! Before }}
<AuLoader @message="Only visible for screenreaders" />

{{! After }}
<AuLoader @hideMessage={{true}}>Only visible for screenreaders</AuLoader>
```

### `@disableMessage`
Disabling the message is not recommended for accessibility reasons. This option was originally added because there was no way to display a visible loading text with the previous implementation. Apps could then disable the internal hidden one, and add a visual loading message themselves. This is now no longer needed since the block content will be used as the visual message by default.

```hbs
{{! Before }}
<AuLoader @disableMessage={{true}} />
<p>Some loading message</p>

{{! After }}
<AuLoader>Some loading message</AuLoader>
```

> If you have a valid reason for not showing any sort of loading message, please open an issue to request this.

### `@padding`
The padding argument was used to align the loader with other whitespace on the page, since the loader is now centered this shouldn't be needed anymore.

It's still possible to add whitespace by using the [spacing util classes](https://appuniversum.github.io/ember-appuniversum/?path=/story/utilities-spacing--page).

### DIFF
--- a/addon/components/au-button.gts
+++ b/addon/components/au-button.gts
@@ -1,6 +1,6 @@
 import Component from '@glimmer/component';
 import AuIcon from './au-icon';
-import AuLoader from './au-loader';
+import { LoadingAnimation } from '../private/components/loading-animation';
 
 const SKINS = [
   'primary',
@@ -118,14 +118,14 @@ export default class AuButton extends Component<AuButtonSignature> {
       {{#if @hideText}}
         {{#if @loading}}
           <span class="au-u-hidden-visually">{{this.loadingMessage}}</span>
-          <AuLoader @padding="small" />
+          <LoadingAnimation />
         {{else}}
           <span class="au-u-hidden-visually">{{yield}}</span>
         {{/if}}
       {{else}}
         {{#if @loading}}
           {{this.loadingMessage}}
-          <AuLoader @padding="small" />
+          <LoadingAnimation />
         {{else}}
           {{yield}}
         {{/if}}
--- a/addon/components/au-loader.gts
+++ b/addon/components/au-loader.gts
@@ -1,10 +1,39 @@
 import Component from '@glimmer/component';
+import { deprecate } from '@ember/debug';
+import { LoadingAnimation } from '../private/components/loading-animation';
+import { modifier } from 'ember-modifier';
+
+const deprecateOldVersion = modifier(function deprecateOldVersion() {
+  deprecate(
+    `[AuLoader] This version of the \`AuLoader\` component is deprecated.
+
+Follow the migration guide to switch to the new version: https://github.com/appuniversum/ember-appuniversum/pull/464
+
+`,
+    false,
+    {
+      id: '@appuniversum/ember-appuniversum.au-loader-visible-by-default',
+      until: '4.0.0',
+      for: '@appuniversum/ember-appuniversum',
+      since: {
+        available: '3.1.0',
+        enabled: '3.1.0',
+      },
+    },
+  );
+});
 
 export interface AuLoaderSignature {
   Args: {
+    inline?: boolean;
+    hideMessage?: boolean;
+    // Deprecated arguments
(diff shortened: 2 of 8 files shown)


tests/dummy/app/templates/application.hbs
Windvis, line 1: This file should be removed before merging, I was testing things out here since the Storybook version didn't update as I would expect.

Windvis:
We also use the `AuLoader` in the [`AuButton` component](https://github.com/appuniversum/ember-appuniversum/blob/9ef3e00ddf0b88f2be3c9fcb7701df0b24f828a7/addon/components/au-button.gts#L128), and I _think_ we don't want to output any loading text there, since the button already does that, but the current PR doesn't support that (in a non-deprecated way). 

We could keep the `@disableMessage` but I'm not sure if we should. I think we want to push the version with a loading message for accessibilty reasons, but maybe there are valid cases where we only want the loading animation. We could go for a private argument that we only use in Appuniversum itself maybe. 🤔.

Another option could be to display the inline version inside the button instead, maybe? It will look a bit different from the current version though.

@brenner-company Any insights?

brenner-company:
I didn't have the time today to give this a look. I'll pick this up on wednesday if that's ok?

brenner-company:
I think both the 'private argument' & 'custom animation (within the button)' are good options. I'm not a fan of the 'inline version' solution because of, as you said, the visual difference.

Could it also be an option to extract the `<div class="au-c-loader__animation" aria-hidden="true"></div>` as a private component (somethin along `addon/private/components/loader-animation`) and use that within the actual loader & button component? Maybe I'm overthinking it 😅

brenner-company (approved):

Windvis:
> I think both the 'private argument' & 'custom animation (within the button)' are good options. I'm not a fan of the 'inline version' solution because of, as you said, the visual difference.

✅ 

> Could it also be an option to extract the <div class="au-c-loader__animation" aria-hidden="true"></div> as a private component (somethin along addon/private/components/loader-animation) and use that within the actual loader & button component? Maybe I'm overthinking it 😅

Yea, that's indeed an option 👍 

The main thing I'm not sure about is, do we want to prevent apps from using the `AuLoader` component without a loading message (so not a non-visual one either). With the current component that's possible. With the new one that won't be. I don't think it's needed, but I might be missing use-cases 😅.

PR #465, opened 2024-02-06
Fix the responsive flex utilities

### LINKED ISSUE #394: Breakpoints work opposite for flex utilities
**Describe the bug**
when using breakpoints in classes like `au-u-1-2@medium`, it will be used when the threshold is *passed* (screen size bigger than `medium`).
for the flex classes like `au-u-flex--wrap@medium`, it will be used when the threshold is *below* (screen size smaller than `medium`).

**To Reproduce**
```
<div class="au-u-flex--wrap@medium"></div>
<div class="au-u-2-3@medium"></div>
```
1) make the screen small. first div will have styles of `au-u-flex--wrap`. Second div will have no styling.
2) Make screen bigger than @medium size: first div will have no styles. Second div will have style of `au-2-3`.

**Expected behavior**
The breakpoints have the same behavior. in point `1)`, both divs have no styles, in point `2)` both have the respective styles.

**Desktop (please complete the following information):**

- OS: Arch Linux
- Browser Firefox
- Version 111.0.1

Windvis:
These were accidentally using the breakpoints as the upper bound instead of lower bound. 
This deprecates the old behavior and adds a global variable that can be set to enable the min-width based queries.

Closes #394

### DIFF
--- a/styles/utilities/_u-flex.scss
+++ b/styles/utilities/_u-flex.scss
@@ -10,6 +10,7 @@
   ========================================================================== */
 
 $au-flex-utilities-responsive: false !default;
+$au-flex-utilities-reverse-responsive-breakpoints: false !default;
 
 /* Utility classes
   ========================================================================== */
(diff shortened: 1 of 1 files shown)


Windvis:
This is breaking for apps that are using the utilities, but we could consider it a bugfix and just mention it in the release notes?

PR #469, opened 2024-02-15
Add an `isOpenInitially` argument to the `AuAccordion` component

### LINKED ISSUE #461: AuAccordion - Add defaultOpen argument
Be able to define a default Accordion state is a standard need. 

Add `defaultOpen: boolean` argument

Need by Lokaal Beslist, cf. https://github.com/lblod/frontend-burgernabije-besluitendatabank/pull/186

Windvis:
This adds a new ~~`defaultOpen`~~ `isOpenInitially` argument to the `AuAccordion` component which can be used to display the accordion in the open state from the start.

It also adds Glint support to component (also to the `AuToolbar` component, since that's one of its dependencies.)

Closes #461

### DIFF
--- a/addon/components/au-accordion.gts
+++ b/addon/components/au-accordion.gts
@@ -18,6 +18,7 @@ export interface AuAccordionSignature {
     buttonLabel?: string;
     iconClosed?: string;
     iconOpen?: string;
+    isOpenInitially?: boolean;
     loading?: boolean;
     reverse?: boolean;
     skin?: 'border';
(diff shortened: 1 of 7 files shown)


Windvis:
@aliokan I've used the `defaultOpen` name since that's how it was described in the issue. I did notice we already have a similar argument for the `AuCard` component but it's called `isOpenInitially`. I think it makes sense to make these somewhat consistent?

So we could rename it here, or add support for `defaultOpen` to the other one as well. (Or just keep it as-is) Any opinions?

aliokan:
@Windvis Thank for the update! :pray: 

As I can see, `AuCard isOpenInitially` is already use in different places: https://github.com/search?q=org%3Alblod%20isOpenInitially&type=code
In order to avoid breaking changes, I would also use `isOpenInitially` in the accordion.

Issue #425, opened 2023-09-12
Provide TS and Glint types


Windvis:
Official support for this in Ember is starting to get better and better so it would be nice if we can support it as a first-class citizen as well.

Some options:
- keep writing js and write .d.ts files ourselves
- keep writing js and use js doc + a build step to extract the d.ts files
- convert to typescript and build the d.ts files that way

Windvis:
I think we should probably just inline the ember-data-table components here, and convert them to .gts. The way we're extending the base addon components is a bad practise anyways, and this will give us more control. It would also allow us to drop the dependency.

We should take a look at all the exposed data table components and deprecate the ones that we don't want to be used as a standalone thing. We can then remove the app-reexports in the next major.

PR #474, opened 2024-02-29
More Glint / TS conversion

Windvis:
This converts some more components:
- `<AuFileCard>`
- `<AuFileUpload>`
- `<AuMainContainer>`
- `<AuMainFooter>`
- `<AuMainHeader>`
- `<AuModalContainer>`
- `<AuNavigationLink>`
- `<AuPanel>`
- `<AuPill>`
- `<AuLabel>`
- `<AuFieldset>`

#425


addon/components/au-file-upload.gts
```diff
@@ -55,36 +89,42 @@ export default class AuFileUpload extends Component {
...
-  *upload(file) {
+  @action
+  upload(file: UploadFile): void | undefined {
+    this.uploadTask.perform(file);
+  }
+
+  uploadTask = task(async (file: UploadFile) => {
     this.resetErrors();
-    let uploadedFile = yield this.uploadFileTask.perform(file);
+    let uploadedFile: number | null = await this.uploadFileTask.perform(file);
```
Windvis, line 99: I _think_ we can type the return type of the  task instead. Not sure why I did this.

Windvis, line 99: Hmm, I just noticed that this can be a `const` and eslint should complain about that. We're running into this issue:

https://github.com/ember-cli/eslint-plugin-ember/issues/2106

So we would have to temporarily add those rules to our eslint config until we switch to the flat config variant, or until eslint-plugin-ember does it for us.

PR #475, opened 2024-03-11
Apply the `@typescript-eslint/eslint-recommended` rules to .gts files 

Windvis:
By default these rules are only applied to a [whitelist of extensions](https://github.com/typescript-eslint/typescript-eslint/blob/c0e3267da71288d6eb9efe790d8aabb0adcb290e/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts#L19), and .gts isn't part of that. To work around that issue we apply the rules ourselves for now.

This fixes an issue we noticed in #474


.eslintrc.js
```diff
@@ -16,11 +21,12 @@ module.exports = {
     // ts files
     {
       files: ['**/*.{gts,ts}'],
-      extends: [
-        'plugin:@typescript-eslint/eslint-recommended',
```
Windvis, line 20: This is already part of the `@typescript-eslint/recommended` config, and we don't really need it in either case, since we include the rules ourselves now.

PR #476, opened 2024-03-11
fix modals replacing existing modal content

Rahien:
- add insertBefore=null so the content of the existing modal isn't replaced (see in-element docs), 
- make backdrop and modal z-index the same, so new modals can be shown over existing modals

the previous behaviour broke opening a modal when another modal was already open because focus-trap tried to access the previous modal's contents which broke the app.


styles/components/_c-modal.scss
```diff
@@ -6,7 +6,7 @@
    ========================================================================== */
 
 $au-modal-z-index: 9999 !default;
-$au-modal-backdrop-z-index: 9998 !default;
+$au-modal-backdrop-z-index: 9999 !default;
```
Windvis, line 9: Question: Does it still make sense to make this configurable after this change? For this to work properly both z-index values need to be the same I think?

Windvis (commented):
Looks good, and seems to work as expected when I try it in the storybook app. 👍

Could you maybe add a simple test to [this file](https://github.com/appuniversum/ember-appuniversum/blob/master/tests/integration/components/au-modal-test.js) that verifies that multiple modal components can be rendered at the same time? Just checking that both are in the DOM should be good enough I think.

styles/components/_c-modal.scss
Rahien: indeed it should be the same. I wanted to keep it around in case others would use this property, but if you prefer i can remove it instead?

PR #481, opened 2024-03-13
[Experimental feature] Add support for using components as icons

piemonkey:
> [!WARNING]  
> This setup is still in an experimental state. We want to try it out in a couple of specific places before considering this ready to use by all the Appuniversum users. We might still break the setup in future minor versions until we announce it as stable.

<hr>

Another PR in the [line](#424) of [PRs](#430) attempting to address the issue of inline SVGs.

This time the approach is to allow passing of arbitrary components to `AuIcon` (and by extension, any component taking an `@icon` argument).

The advantages of this method are that this is very flexible to the user. Also, by overloading `@icon` it avoids confusion that could be caused by adding a new argument, as only one of `string` or `Component` can be passed.

A minor downside is that the user can potentially pass a component that could cause rendering issues.

A risk is that in order to pass arguments to the component, for example; to pass an `AuIcon` as a custom component, you need an argument to specify which icon to use; you must use the `component` helper. In this way it is easy to write code that embroider can't statically analyse.

### DIFF
--- a/addon/components/au-accordion.gts
+++ b/addon/components/au-accordion.gts
@@ -5,7 +5,7 @@ import { tracked } from '@glimmer/tracking';
 import { modifier } from 'ember-modifier';
 import AuButton from './au-button';
 import AuContent from './au-content';
-import AuIcon from './au-icon';
+import AuIcon, { type AuIconSignature } from './au-icon';
 import AuLoader from './au-loader';
 import AuToolbar from './au-toolbar';
 
@@ -16,8 +16,8 @@ const autofocus = modifier(function autofocus(element: HTMLElement) {
 export interface AuAccordionSignature {
   Args: {
     buttonLabel?: string;
-    iconClosed?: string;
-    iconOpen?: string;
+    iconClosed?: AuIconSignature['Args']['icon'];
+    iconOpen?: AuIconSignature['Args']['icon'];
     isOpenInitially?: boolean;
     loading?: boolean;
     reverse?: boolean;
--- a/addon/components/au-alert.gts
+++ b/addon/components/au-alert.gts
@@ -2,12 +2,12 @@ import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
-import AuIcon from './au-icon';
+import AuIcon, { type AuIconSignature } from './au-icon';
 
 export interface AuAlertSignature {
   Args: {
     closable?: boolean;
-    icon?: string;
+    icon?: AuIconSignature['Args']['icon'];
     iconVisible?: boolean;
     onClose?: () => void;
     size?: 'tiny' | 'small';
--- a/addon/components/au-icon.gts
+++ b/addon/components/au-icon.gts
@@ -7,10 +7,10 @@ export interface AuIconSignature {
     alignment?: 'left' | 'right';
     // TODO: We should deprecate the non-boolean versions since there is no reason to support them
     ariaHidden?: boolean | 'true' | 'false';
-    icon: string | ComponentLike<Record<string, never>>;
+    icon: string | ComponentLike<{ Element: Element }>;
     size?: 'large';
   };
-  Element: SVGSVGElement;
+  Element: Element;
 }
 
 export default class AuIcon extends Component<AuIconSignature> {
@@ -47,10 +47,10 @@ export default class AuIcon extends Component<AuIconSignature> {
   <template>
     {{#if this.iconComponent}}
       {{#let this.iconComponent as |Icon|}}
-        {{! @glint-expect-error: glint doesn't like us setting attributes on the passed component }}
         <Icon
           class="au-c-icon {{this.alignment}} {{this.size}}"
           aria-hidden={{this.ariaHidden}}
+          ...attributes
         />
       {{/let}}
     {{else}}
--- a/addon/components/au-icon.gts
+++ b/addon/components/au-icon.gts
@@ -1,12 +1,13 @@
 import { getOwner } from '@ember/owner';
 import Component from '@glimmer/component';
+import { type ComponentLike } from '@glint/template';
 
 export interface AuIconSignature {
   Args: {
     alignment?: 'left' | 'right';
     // TODO: We should deprecate the non-boolean versions since there is no reason to support them
     ariaHidden?: boolean | 'true' | 'false';
-    icon: string;
+    icon: string | ComponentLike<Record<string, never>>;
     size?: 'large';
   };
   Element: SVGSVGElement;
@@ -39,16 +40,30 @@ export default class AuIcon extends Component<AuIconSignature> {
     }
   }
 
+  get iconComponent() {
+    return typeof this.args.icon !== 'string' && this.args.icon;
+  }
+
   <template>
-    <svg
-      role="img"
-      class="au-c-icon au-c-icon--{{@icon}} {{this.alignment}} {{this.size}}"
-      aria-hidden={{this.ariaHidden}}
-      ...attributes
-    >
-      <use
-        xlink:href="{{this.iconPrefix}}@appuniversum/ember-appuniversum/appuniversum-symbolset.svg#{{@icon}}"
-      ></use>
-    </svg>
+    {{#if this.iconComponent}}
+      {{#let this.iconComponent as |Icon|}}
+        {{! @glint-expect-error: glint doesn't like us setting attributes on the passed component }}
+        <Icon
+          class="au-c-icon {{this.alignment}} {{this.size}}"
+          aria-hidden={{this.ariaHidden}}
+        />
+      {{/let}}
+    {{else}}
+      <svg
+        role="img"
+        class="au-c-icon au-c-icon--{{@icon}} {{this.alignment}} {{this.size}}"
+        aria-hidden={{this.ariaHidden}}
+        ...attributes
+      >
+        <use
+          xlink:href="{{this.iconPrefix}}@appuniversum/ember-appuniversum/appuniversum-symbolset.svg#{{@icon}}"
+        ></use>
+      </svg>
+    {{/if}}
   </template>
 }
--- a/addon/components/au-badge.gts
+++ b/addon/components/au-badge.gts
@@ -1,9 +1,9 @@
 import Component from '@glimmer/component';
-import AuIcon from './au-icon';
+import AuIcon, { type AuIconSignature } from './au-icon';
 
 export interface AuBadgeSignature {
   Args: {
-    icon?: string;
+    icon?: AuIconSignature['Args']['icon'];
     iconVisible?: boolean;
     number?: number;
     size?: 'small';
--- a/addon/components/au-button.gts
+++ b/addon/components/au-button.gts
@@ -1,5 +1,5 @@
 import Component from '@glimmer/component';
-import AuIcon from './au-icon';
+import AuIcon, { type AuIconSignature } from './au-icon';
 import { LoadingAnimation } from '../private/components/loading-animation';
 
 const SKINS = [
@@ -15,7 +15,7 @@ export interface AuButtonSignature {
     alert?: boolean;
     disabled?: boolean;
     hideText?: boolean;
-    icon?: string;
+    icon?: AuIconSignature['Args']['icon'];
     iconAlignment?: 'left' | 'right';
     loading?: boolean;
     loadingMessage?: string;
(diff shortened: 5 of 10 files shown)


addon/components/au-icon.gts
```diff
@@ -6,7 +6,7 @@ export interface AuIconSignature {
     alignment?: 'left' | 'right';
     // TODO: We should deprecate the non-boolean versions since there is no reason to support them
     ariaHidden?: boolean | 'true' | 'false';
-    icon: string;
+    icon: string | Component<Record<string, never>>;
```
Windvis, line 9: I think you want to use the `ComponentLike` type here: https://typed-ember.gitbook.io/glint/using-glint/glint-types#componentlike-helperlike-and-modifierlike

addon/components/au-icon.gts
```diff
@@ -39,16 +39,30 @@ export default class AuIcon extends Component<AuIconSignature> {
...
-      <use
-        xlink:href="{{this.iconPrefix}}@appuniversum/ember-appuniversum/appuniversum-symbolset.svg#{{@icon}}"
-      ></use>
-    </svg>
+    {{#if this.iconComponent}}
+      {{#let this.iconComponent as |IconComponent|}}
+        {{! @glint-expect-error: glint seems to think that this is not a valid component type  }}
+        <IconComponent
+          class="au-c-icon {{this.alignment}} {{this.size}}"
+          aria-hidden={{this.ariaHidden}}
```
Windvis, line 52: I'm not sure if we should pass any extra attributes to this component actually. This feels like we leak implementation details of the `AuIcon` component into the custom ones? I think we should just render the custom component as-is for now.

addon/components/au-icon.gts
```diff
@@ -39,16 +39,30 @@ export default class AuIcon extends Component<AuIconSignature> {
...
-      aria-hidden={{this.ariaHidden}}
-      ...attributes
-    >
-      <use
-        xlink:href="{{this.iconPrefix}}@appuniversum/ember-appuniversum/appuniversum-symbolset.svg#{{@icon}}"
-      ></use>
-    </svg>
+    {{#if this.iconComponent}}
+      {{#let this.iconComponent as |IconComponent|}}
+        {{! @glint-expect-error: glint seems to think that this is not a valid component type  }}
```
Windvis, line 49: I think this might be fixed by using the `ComponentLike` type.

addon/components/au-icon.gts
```diff
@@ -39,16 +39,30 @@ export default class AuIcon extends Component<AuIconSignature> {
...
-      class="au-c-icon au-c-icon--{{@icon}} {{this.alignment}} {{this.size}}"
-      aria-hidden={{this.ariaHidden}}
-      ...attributes
-    >
-      <use
-        xlink:href="{{this.iconPrefix}}@appuniversum/ember-appuniversum/appuniversum-symbolset.svg#{{@icon}}"
-      ></use>
-    </svg>
+    {{#if this.iconComponent}}
+      {{#let this.iconComponent as |IconComponent|}}
```
Windvis, line 48: Nitpick: I would name it `Icon` if we're aliasing things to make it nicer looking 😄.
```suggestion
      {{#let this.iconComponent as |Icon|}}
```

Windvis (commented):
Thanks for working on this!

Some quick feedback. I'll try to take a better look tomorrow.

addon/components/au-icon.gts
piemonkey: Ah, didn't know that existed, thanks.

addon/components/au-icon.gts
piemonkey: Hmm, I was wondering about that but also wanted to take care of accessibility, as well as preventing the need to manually match the styling and sizing. I guess we could just document the classes/rules to apply to match the existing implementation and advice on accessibility? Or we could put the Icon in a div, but I worry that will have other side-effects.

addon/components/au-icon.gts
piemonkey: Hmm, it's fixed with `ComponentLike` if I don't pass any attributes.

Windvis:
Thinking about this some more. I don't think this will fix the problem in the frontend-embeddable-notule-editor project, since we also use the `AuIcon` component internally with some "hardcoded icons" and those would still look at the symbolset version. Making those configurable would be strange, since the icon is just an implementation detail of the component.

So while I still think accepting component for `@icon` is a good improvement (to allow custom icons) it still won't fix the problem you are seeing in the project. 🤔

addon/components/au-icon.gts
Windvis: Hmm, yea, maybe. I just wonder if there might be cases where we don't want these attributes. I'm probably overthinking and this might be the safest option. 

I do still wonder why we would ever set aria-hidden to something else than "true" though, I'm not sure why this was ever an argument in the first place. Do you think it makes sense to always set this to true for the icon components, since it's an opt-in?

addon/components/au-icon.gts
```diff
@@ -39,16 +40,30 @@ export default class AuIcon extends Component<AuIconSignature> {
...
-        xlink:href="{{this.iconPrefix}}@appuniversum/ember-appuniversum/appuniversum-symbolset.svg#{{@icon}}"
-      ></use>
-    </svg>
+    {{#if this.iconComponent}}
+      {{#let this.iconComponent as |Icon|}}
+        {{! @glint-expect-error: glint doesn't like us setting attributes on the passed component }}
+        <Icon
+          class="au-c-icon {{this.alignment}} {{this.size}}"
+          aria-hidden={{this.ariaHidden}}
+        />
```
Windvis, line 54: I would also spread the incoming attributes on this component. That way consumers can pass extra attributes to the AuIcon, and they would still be passed down.

```suggestion
          ...attributes
        />
```

```hbs
<AuIcon @icon={{SomeIconComponent}} class="hello" />
```

I'm not sure how that translates into the component signature types though. In theory the "Element" of the signature is then dynamic. 'SVGSVGElement' if `@icon` is a string, and `Element` otherwise, since we don't know what specific element the passed in component uses. I'm not sure if that is even possible, but I'm not very experienced in TS.

stories/4-utilities/Icons.stories.mdx
Windvis, line 1: I wouldn't ship these docs yet, or mark the icon part as an experimental feature that can still change in future minor versions so people know the risks of using them.

stories/5-components/Brand/AuIcon.stories.js
```diff
@@ -42,6 +42,17 @@ const Template = (args) => ({
   context: args,
 });
 
+const CustomIconTemplate = (args) => ({
+  template: hbs`
+    <AuIcon
+      @icon={{component 'au-icon' icon=this.icon}}
```
Windvis, line 48: This example feels a bit silly since there is no real reason to do this 😄. I wouldn't add the example yet. I'll probably add something as part of #483 or when we consider this feature no longer experimental.

addon/components/au-alert.gts
```diff
@@ -2,12 +2,12 @@ import { on } from '@ember/modifier';
...
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
-import AuIcon from './au-icon';
+import AuIcon, { type AuIconSignature } from './au-icon';
 
 export interface AuAlertSignature {
   Args: {
     closable?: boolean;
-    icon?: string;
+    icon?: AuIconSignature['Args']['icon'];
```
Windvis, line 10: I think we need similar changes in a couple of other components that also accept configurable icons:

- au-accordion
- au-dropdown
- au-link-external

addon/components/au-icon.gts
Windvis: Yea, I think the type isn't 100% correct, and an "Element" [is needed](https://github.com/appuniversum/ember-appuniversum/pull/481#discussion_r1524600624).

addon/components/au-icon.gts
piemonkey: So you're happy with leaving these attributes? I'm wondering if we should give the option to opt out of these attributes being applied.

addon/components/au-icon.gts
```diff
@@ -1,12 +1,13 @@
...
 import Component from '@glimmer/component';
+import { type ComponentLike } from '@glint/template';
 
 export interface AuIconSignature {
   Args: {
     alignment?: 'left' | 'right';
     // TODO: We should deprecate the non-boolean versions since there is no reason to support them
     ariaHidden?: boolean | 'true' | 'false';
-    icon: string;
+    icon: string | ComponentLike<Record<string, never>>;
```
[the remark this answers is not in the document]
piemonkey: Ah, yes. The `Record<string, never>` was based on my original use of `Component<Args>` as the type, so I needed to replace the default `unknown` with something that represented no arguments. I'll update this and pass `...attributes` to `Icon`.

addon/components/au-icon.gts
piemonkey: I (as demonstrated in this PR) don't know anything about the new glint types. We could make `AuIcon` generic, taking the Element type as the type argument, but that would only be meaningful if there was a way to specify that in the `*.gts` template syntax. The `SuperTable` example in [the docs](https://typed-ember.gitbook.io/glint/environments/ember/component-signatures) actually is generic, but there's no example of it being used. We can probably stick to `Element` for now.

stories/4-utilities/Icons.stories.mdx
piemonkey: I'll move the docs to a separate PR. Unless there's some existing way to mark them as experimental?

stories/5-components/Brand/AuIcon.stories.js
piemonkey: True, but I thought it was better than no example. The alternative was to add a component just for this example, but I thought that might be even more confusing...

addon/components/au-alert.gts
piemonkey: Huh, I thought I'd searched for all of these, I'll change those and do another search through for any I missed.

Windvis: ah, au-dropdown is still a .gjs file, so ignore that one. Maybe I converted the others after you created your PR, not sure 😄.

stories/5-components/Brand/AuIcon.stories.js
Windvis: Yea, I would have gone for a custom example component in the dummy app. Maybe we are just confused by different things 😄. Anyways, I would split it off or remove it for now.

stories/4-utilities/Icons.stories.mdx
Windvis: > Unless there's some existing way to mark them as experimental?

I don't think so, I was just thinking about adding an alert message or something to the docs, but splitting it to a separate PR is better 👍.

addon/components/au-icon.gts
Windvis: > So you're happy with leaving these attributes?

Yea, it seems like the safest option to start with 👍.

> I'm wondering if we should give the option to opt out of these attributes being applied.

Well, yea that's what I was originally thinking, but that's probably not needed. I suspect that most consumers will always use one of the bundled icon components, or a custom svg component. I don't expect that people will want to fully replace the AuIcon behavior, but I could be wrong.

I think this is one of those cases that we can't predict, and only encounter once we start using this system in our apps. So I would keep it simple for now.

addon/components/au-accordion.gts
```diff
@@ -107,15 +110,15 @@ export default class AuAccordion extends Component<AuAccordionSignature> {
               @alignment="left"
               @size="large"
               @ariaHidden={{true}}
-              data-test-accordion-icon-open={{this.iconOpen}}
+              data-test-accordion-icon-open={{iconTestName this.iconOpen}}
```
Windvis, line 113: Hmmm, I think we should just remove the attribute and corresponding test to be honest. Adding an extra helper call simply to test something seems unneeded, and I'm not convinced it was a good test in the first place (it probably makes more sense to use visual regression tests for something like this).

Also, was this helper wrapper needed? I would have expected that Ember would toString the value since it's an attribute. Or was this for type reasons maybe?

Windvis: I can also do that in master already, after which you can rebase? It seems unrelated to this PR, I just noticed it here. Thoughts?

piemonkey: Yeah, makes sense to me to remove it separately and rebase on that.

piemonkey: ...and done.

Windvis (approved):

Issue #482, opened 2024-03-14
Appuniversum icon improvements


Windvis:
At the moment Appuniversum bundles a [list of icons](https://github.com/appuniversum/ember-appuniversum/tree/654711add804e8e2de2a5b23b22976690a9c4b90/public/icons) into a svg-symbols file which is also stored in the public folder and which is then referenced in `AuIcon` component.

This works, but has some downsides:
- no support for custom icons, which means icons need to be added here, even if they are very specific for a single project
- All the icons are bundled, even if only a few are used by a project
- If the app is embedded in a different domain, the browser might refuse to load the referenced symbolset, so something like [svgxuse](https://github.com/Keyamoon/svgxuse) needs to be used as a workaround.

An potential solution for these issues is to use icon components, instead of the string based setup we have now. Each .svg file could be converted to a .gts file at build time, and imported like any other component.

The benefits:
- Since they're components, Embroidered apps will strip out the unused ones so only the used icons will be part of the build
- No reference to a symbolset file, so no issues with cross-domain requests

For this to work, each component that now accepts a string for the `@icon` argument, would also need to support a component (POC in #481 ).

We would also need to refactor our internal usage of the `AuIcon` component, and pass the components instead of the string.

**Open questions:**
- Do we want to deprecate the string based system? If we don't any usage of the symbolset will trigger the download resulting in all the icons being downloaded again)
- Do we want to expose these components in Ember's global component namespace, or do we only expose the export paths and require that consumers import where needed? (In non-gjs/gts files this might be cumbersome, but I think we're at a point where using .gjs is in a good place)

piemonkey:
> If the app is embedded in a different domain, the browser might refuse to load the referenced symbolset, so something like [svgxuse](https://github.com/Keyamoon/svgxuse) needs to be used as a workaround.

In fact, while the svgxuse workaround does work if you're hosting the ember app that is embedded in another domain (for example in an iframe), as svgxuse is still able to request the symbolset. If the app is designed to be bundled within another project, then the symbolset is no longer available and this approach fails. This is the case for [the embeddable-say-editor](https://github.com/lblod/frontend-embeddable-notule-editor/) project which can now be installed through npm and in order to work requires overriding `AuIcon` with a custom version. This workaround only works for ember-appuniversum <=2.15.

abeforgit:
[...]

PR #483, opened 2024-03-19
[Experimental] Add icon components

Windvis:
This adds icon components for each .svg file in the `/public/icons` folder. We don't register them in the global component namespace, to use them you need to import them in your JS files and pass them through the template context. This a lot more intuitive in the .gjs world, but it still works in loose-mode as well, but it is more verbose.

This PR isn't very useful by itself, but together with #481 it can be used to replace the string based icon setup.

> [!WARNING]  
> This setup is still in an experimental state. We want to try it out in a couple of specific places before considering this ready to use by all the Appuniversum users. We might still break the setup in future minor versions until we announce it as stable.

<details>
<summary>Usage examples</summary>

(These examples assume #481 is merged)
### Polaris

```gjs
// some-delete-button.gjs
import { BinIcon } from '@appuniversum/ember-appuniversum/components/icons/bin';
import AuButton from '@appuniversum/ember-appuniversum/components/au-button';

<template>
  <AuButton
    @icon={{BinIcon}}
    @alert={{true}}
  >
    Delete something
  </AuButton>
</template>
```

### Octane
```js
// some-delete-button.js
import { BinIcon } from '@appuniversum/ember-appuniversum/components/icons/bin';

export default class SomeDeleteButton extends Component {
  BinIcon = BinIcon;
}
```
```hbs
{{! some-delete-button.hbs}}
<AuButton
  @icon={{this.BinIcon}}
  @alert={{true}}
>
  Delete something
</AuButton>
```
</details>

Part of the solution for #482


lib/generate-icon-components.mjs
Windvis, line 1: This file needs to be npmignored, since we only need to run it prepublish.

lib/generate-icon-components.mjs
```diff
@@ -0,0 +1,46 @@
...
+async function generateComponent(iconName) {
+  const componentName = pascalCase(iconName);
+
+  const iconContent = await readFile(join(ICON_FOLDER, iconName + '.svg'));
+  const iconContentWithAttributes = iconContent.toString().replace('>', ' ...attributes>'); // We assume the first closing element is the svg element
+
+  const component = `// THIS FILE IS GENERATED. ANY CHANGES TO THIS FILE WILL BE LOST.
+import type { TOC } from '@ember/component/template-only';
+
+export interface ${componentName}IconSignature {
```
piemonkey, line 31: Since the interface is static, this could live separately and be imported into each file, which would simplify the generated file slightly.

Windvis: Yea, but since all the files are generated it feels a bit strange to store a common type file somewhere, from which they would import. I don't really see an added value, other than DRYing up the generated code a tiny bit, but we would edit the generation script if we would want to make changes anyways, so no real benefit there either. I would prefer to keep it like this for simplicity sake.

piemonkey: It was just a similar strange feeling with generating something that is static, but I don't see a strong reason to pick one feeling over the other.

Windvis:
I forgot to squash merge.. my bad..

Windvis:
Reverted this change and reopened as #487

PR #486, opened 2024-03-21
Use the icon components internally

Windvis:
This is a good experiment to see if everything works, and it allows projects to skip the svg-symbolset download if they also refactor their own code.

Builds on #483 and is part of #482

### DIFF
--- a/addon/components/au-accordion.gts
+++ b/addon/components/au-accordion.gts
@@ -8,6 +8,8 @@ import AuContent from './au-content';
 import AuIcon, { type AuIconSignature } from './au-icon';
 import AuLoader from './au-loader';
 import AuToolbar from './au-toolbar';
+import { NavDownIcon } from './icons/nav-down';
+import { NavRightIcon } from './icons/nav-right';
 
 const autofocus = modifier(function autofocus(element: HTMLElement) {
   element.focus();
@@ -48,7 +50,7 @@ export default class AuAccordion extends Component<AuAccordionSignature> {
     if (this.args.iconOpen) {
       return this.args.iconOpen;
     } else {
-      return 'nav-down';
+      return NavDownIcon;
     }
   }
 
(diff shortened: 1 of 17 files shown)


Windvis:
@piemonkey (CC @abeforgit) this should cover everything to unblock the embeddable I think.

I'm not familiar with the project, but maybe it might be interesting to `yalc` this PR and try it out in the app? Or do you prefer merging and releasing as an experimental feature?

Windvis:
> I didn't realise that you'd gone for the build-on-package approach in the end.

Yea, it didn't feel right and I consider this a temporary setup anyways since the script will be converted to a rollup plugin once we switch to the v2 addon setup.

> I did have to use yalc in the end due to the combination of npm lifecycle scripts and git: dependencies being broken.

While looking into this last week I came across a comment from an npm maintainer who seemed to imply that he considers using github dependencies a very bad practice, so that might explain why things like this don't get fixed 😄. 

> After updating our use of components to pass icon components instead of strings, this version does indeed fix the problem.

Thanks for testing 🙏. I'll merge and tag a release today.

Windvis:
>  I did have to use yalc in the end due to the combination of npm lifecycle scripts and git: dependencies being broken.

Wait, was the issue related to missing dependencies? It seems npm only installs devDeps for git dependencies if there is a prepare script. We're using `prepack` now (someone in the Discord recommended this) so that might explain it.

> NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.


In any case, I'm not sure if we should actually support using Appuniversum as a Git dependency to be honest, so I'll just leave it as-is for now.

piemonkey:
> > I did have to use yalc in the end due to the combination of npm lifecycle scripts and git: dependencies being broken.
> 
> Wait, was the issue related to missing dependencies? It seems npm only installs devDeps for git dependencies if there is a prepare script. We're using `prepack` now (someone in the Discord recommended this) so that might explain it.

I forget for certain now, but I think it was not generating the components at all. I don't think it's a thing that needs to be fixed as I think trying to find a solution that works for that and in all the other possible configurations is either impossible or very hard. I just mentioned it to potentially avoid others wasting time with it in the future.

abeforgit:
chiming in with: npm git dependencies -> "just don't"

PR #488, opened 2024-03-27
Remove the fill attribute from all icons

Windvis:
The fill attribute prevents the icon from receiving the proper coloring. This wasn't an issue in the symbolset version because svg-symbols strips the attribute.

Part of #482


public/icons/github.svg
```diff
@@ -1,5 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" viewBox="0 0 24 24">
```
Windvis, line 1: This one wasn't really a problem since we also use `currentColor` but it seemed better to remove it for consistency.

Windvis:
Released in v3.4.1

PR #492, opened 2024-04-24
Resolve some issues with `babel-plugin-ember-template-compilation` v2.2.2+

### LINKED ISSUE #491: `babel-plugin-ember-template-compilation` v2.2.2 incompatibilities
https://github.com/emberjs/babel-plugin-ember-template-compilation/releases/tag/v2.2.2 includes some fixes for bugs we might actually be depending on by accident.

The error message looks like this: `could not find module ember-template-compiler`

A quick workaround is to pin `babel-plugin-ember-template-compilation` to v2.2.1 using overrides until we can debug and fix the issue here.

```json
  "overrides": {
    "babel-plugin-ember-template-compilation": "2.2.1"
  },
```

Related issue: https://github.com/emberjs/babel-plugin-ember-template-compilation/issues/36

Windvis:
The v2.2.2 release breaks our ember-template-imports v3 based setup.

We need to update to v4 to fix those and also fix the scope issues we accidentally depended on.

~~Draft since it seems to break the storybook setup. We probably need to update that first?~~ Not sure what changed or what the issue was, but storybook seems to work now.

Fixes #491 

TODO:
- [x] clean up the PR, lots of changes got in to test things but those aren't needed anymore.


package.json
```diff
@@ -73,7 +73,8 @@
   "overrides": {
     "@glimmer/validator": "0.84.3",
     "@glimmer/util": "0.84.3",
-    "ember-source": "$ember-source"
+    "ember-source": "$ember-source",
+    "ember-cli-htmlbars": "^6.3.0"
```
Windvis, line 77: This was added to fix an issue with ember canary, but that issue will be fixed upstream so we can remove the override.

Windvis:
(Canary will fail until the fix is merged in ember-source, but we'll just ignore that since it will resolve itself eventually)

Issue #478, opened 2024-03-11
Add a Tooltips feature


Windvis:
Webuniversum has a [tooltip feature](https://overheid.vlaanderen.be/webuniversum/v3/documentation/js-components/vl-ui-tooltip) that could also be useful for Appuniversum projects.

The details of how this should be implemented still need to be refined. Kaleidos already uses a custom version of this, so they would benefit from adding this to Appuniversum.

Windvis:
It seems there is a fairly recent native "popover" feature in browsers that we could look into: 

- https://codepen.io/melsumner/details/VwNmYLY
- https://developer.mozilla.org/en-US/docs/Web/API/Popover_API

Saskapallayova:
I've noticed already 2 separate LBLOD platforms (Validation tool & LMB) use tooltips with different UIs, and I'm pretty sure GN also uses a tooltip somewhere I just can't find it :D 

I'd say we have grounds to add this to AU now, so we can all use the same UI for them? 

Examples of inconsistencies throughout the platforms: 
<img width="523" alt="Screenshot 2024-10-14 at 10 48 24" src="https://github.com/user-attachments/assets/fd9e7551-a3bf-4f01-a63b-0add42de6abf">
<img width="265" alt="Screenshot 2024-10-14 at 10 48 36" src="https://github.com/user-attachments/assets/1fd9133e-ec2c-4748-8297-f97fe85b6554">
<img width="258" alt="Screenshot 2024-10-14 at 10 48 48" src="https://github.com/user-attachments/assets/a8080224-c304-4ac0-a234-36a4e41b6f64">

PR #493, opened 2024-06-12
Tooltip component

snenenenenenene:
As discussed during the demo last sprint I would work on adding the tooltip component used in the validation monitoring tool to the ember appuniverse library.

Part of [BNB-675](https://binnenland.atlassian.net/browse/BNB-675)

### DIFF
--- a/addon/components/au-tooltip.gts
+++ b/addon/components/au-tooltip.gts
@@ -1,0 +1,21 @@
+import Component from '@glimmer/component';
+
+export interface AuTooltipSignature {
+  Args: {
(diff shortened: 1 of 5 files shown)


Windvis:
@snenenenenenene I checked the Jira issue but it's not clear for me yet what you are exactly creating. Does it match the description of this issue? https://github.com/appuniversum/ember-appuniversum/issues/478 

Ideally we match the Webuniversum setup. If it's something else I would keep it as a custom thing in the project you are working on :sweat_smile:.

PR #494, opened 2024-07-09
More Glint / TS conversion

Windvis:
Converts:
- `AuDatePicker`
- `{{au-inputmask}}`
- `{{floating-ui}}`
- `AuDropdown`
- `AuTimePicker`

Part of #425


addon/components/au-date-picker.gts
```diff
@@ -11,29 +10,40 @@ import { guidFor } from '@ember/object/internals';
...
-  parse: function (value = '', createDate) {
-    let dateRegex = /^(\d{1,2})-(\d{1,2})-(\d{4})$/;
+export interface AuDatePickerSignature {
+  Args: {
+    alignment?: 'top';
+    adapter?: Adapter;
+    buttonLabel?: string;
+    disabled?: boolean;
+    error?: boolean;
+    'first-day'?: string;
```
Windvis, line 29: We should probably support a `firstDay` argument as well, not sure why we went with a dasherized name in the first place.

PR #498, opened 2024-07-11
Add types for the data-table components

Windvis:
This copies over the ember-data-table .js code so we no longer need the addon itself. We were already overriding all the templates. This will make things easier to maintain and improve in the future.

Linting errors are silenced with inline ignore comments. They will be fixed once we convert the code to .gts.

This is a good base to start the .gts conversion from.

TODO:

- [x] Copy over the tests (and tweak them to match our version of the components)
- [ ] Add types for the classic components

Out of scope:
- [ ] ~~Convert the components to .gts~~
- [ ] ~~deprecate things that we consider private and don't want apps to use directly~~


package.json
```diff
@@ -89,8 +89,8 @@
     "ember-auto-import": "^2.6.3",
     "ember-cli-babel": "^8.2.0",
     "ember-cli-htmlbars": "^6.3.0",
+    "ember-composable-helpers": "^5.0.0",
```
Windvis, line 92: It seems we were using this in the data table template overrides, but we weren't depending on the addon itself. This only worked because ember-data-table brought it in.

Windvis:
Hmmm, I'm not sure if this is possible in a backwards compatible way. Some ember-data-table components mutate arguments, which is no longer possible with Glimmer components. Consumers would have to pass in an action, which isn't a huge issue, but it requires a code change.

It might be possible to make this an opt-in breaking change by using the embroider macros, and only expose the new implementation if the consuming app opted in.

Maybe we should settle on providing types for the classic ember components instead for now.

Windvis:
Closing since it's essentially the same as #525. I'll reopen it once I start on the actual type work.

PR #508, opened 2024-09-05
Accessibility update / Tested for frontend-subsidiepunt

Dietr:
Accessibility fixes:
- Fix AuDataTable sortable header color contrast
- Fix secondary helptext color contrast
- Remove aria-controls from aria-live region: not a valid property

Documentation update:
- fix dropdown example: role=menu should not be on the dropdown button

Additions:
- AuMainHeader: add override on the contact link text. Certain products apply a hack to put a different link which breaks the semantics of this component

### DIFF
--- a/addon/components/au-main-header.gts
+++ b/addon/components/au-main-header.gts
@@ -62,7 +62,7 @@ export default class AuMainHeader extends Component<AuMainHeaderSignature> {
         </a>
       </div>
       <nav
-        aria-label="{{this.navigationAriaLabel}}"
+        aria-label={{this.navigationAriaLabel}}
         class="au-c-main-header__actions"
       >
         <ul class="au-c-list-horizontal">
--- a/addon/components/au-main-header.gts
+++ b/addon/components/au-main-header.gts
@@ -13,6 +13,7 @@ export interface AuMainHeaderSignature {
     appTitle: string;
     brandLink?: string;
     contactRoute?: string;
+    contactLabel?: string;
     homeRoute?: string;
   };
   Blocks: {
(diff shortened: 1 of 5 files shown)


stories/5-components/Navigation/AuDropdown.stories.js
```diff
@@ -57,20 +58,22 @@ const Template = (args) => ({
...
-    </AuDropdown>`,
+    </AuDropdown>
+    {{! template-lint-enable require-context-role }}
+  `,
   context: args,
 });
 
 const TemplateSeparator = (args) => ({
   template: hbs`
+    {{! template-lint-disable require-context-role }}
```
Windvis, line 76: It might be a good idea to yield button and link components that are pre-wired to do the right thing. Apps will also have to disable this rule as well which isn't ideal.

I'll create an issue so I don't forget to look at it.

Windvis (approved):

addon/components/au-main-header.gts
```diff
@@ -49,7 +55,7 @@ export default class AuMainHeader extends Component<AuMainHeaderSignature> {
           Naar de hoofdinhoud
         </a>
       </div>
-      <nav class="au-c-main-header__actions">
+      <nav aria-label="Informatie en instellingen" class="au-c-main-header__actions">
```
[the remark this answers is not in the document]
Dietr: Yes, might indeed be better. I'll just provide a default.

stories/5-components/Navigation/AuDropdown.stories.js
Dietr: Sure, that would be a more robust solution but a bit out of scope of my current assignment.

addon/components/au-main-header.gts
```diff
@@ -56,7 +62,7 @@ export default class AuMainHeader extends Component<AuMainHeaderSignature> {
         </a>
       </div>
       <nav
-        aria-label="Informatie en instellingen"
+        aria-label="{{this.navigationAriaLabel}}"
```
Windvis, line 65: Quotes aren't needed since it's already a string.

```suggestion
        aria-label={{this.navigationAriaLabel}}
```

Windvis (approved):

PR #512, opened 2024-11-05
Add support for ember-concurrency v4 and ember-power-select v8

### LINKED ISSUE #507: Support ember-concurrency v4
It's now a v2 addon, which requires a bit more work setting them up. Apps might still want to start using this new version so we should try to support it, without dropping support for version v3 since we don't want to create a new major release yet.

We should do some node version checking to conditionally do the required work based on the installed version.

Installation guides:
- http://ember-concurrency.com/docs/installation/

Windvis:
Closes #507

### DIFF
--- a/tests/dummy/config/ember-try.js
+++ b/tests/dummy/config/ember-try.js
@@ -40,6 +40,18 @@ module.exports = async function () {
       },
       embroiderSafe(),
       embroiderOptimized(),
+      {
+        name: 'ember-concurrency-v3',
+        npm: {
+          dependencies: {
+            'ember-concurrency': '^3.0.0',
+          },
+          devDependencies: {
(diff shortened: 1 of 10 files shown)


tests/dummy/config/ember-try.js
```diff
@@ -40,6 +40,18 @@ module.exports = async function () {
...
       embroiderSafe(),
       embroiderOptimized(),
+      {
+        name: 'ember-concurrency-v3',
+        npm: {
+          dependencies: {
+            'ember-concurrency': '^3.0.0',
+          },
+          devDependencies: {
+            'ember-basic-dropdown': null,
```
Windvis, line 50: We only installed this because ember-power-select v8 has it listed as a peerDependency. In v7 it's bundled as a dependency.

PR #523, opened 2024-11-07
Update to TypeScript 5.6.3

Windvis:
Minimal reproduction of an eslint issue when updating to TypeScript 5.6.3.

Current config that breaks:
```js
import ts from 'typescript-eslint';
import ember from 'eslint-plugin-ember/recommended';

const parserOptions = {
  esm: {
    ts: {
      projectService: true,
      tsconfigRootDir: import.meta.dirname,
    },
  },
};

export default ts.config(
  {
    ignores: ['**/*.js', '**/*.mjs'],
  },
  {
    name: 'ts',
    files: ['**/*.ts'],
    languageOptions: {
      // Uncommenting this fixes the issue so it seems to be related to switching parsers for different kinds of typed files?
      // parser: ember.parser,
      parserOptions: parserOptions.esm.ts,
    },
    extends: [...ts.configs.recommendedTypeChecked],
    // extends: [...ts.configs.recommended], // also fails
  },
  {
    name: 'gts',
    files: ['**/*.gts'],
    languageOptions: {
      parser: ember.parser,
      parserOptions: parserOptions.esm.ts,
    },
    // These don't really matter, commented or not, it fails
    // extends: [...ts.configs.recommendedTypeChecked, ember.configs.gts],
  },
);
```

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,45 +1,8 @@
-/**
- * Debugging:
- *   https://eslint.org/docs/latest/use/configure/debug
- *  ----------------------------------------------------
- *
- *   Print a file's calculated configuration
- *
- *     npx eslint --print-config path/to/file.js
- *
- *   Inspecting the config
- *
- *     npx eslint --inspect-config
- *
- */
-
-import globals from 'globals';
-import js from '@eslint/js';
-
 import ts from 'typescript-eslint';
-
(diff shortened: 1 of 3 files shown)


Windvis:
Closing since we'll just switch to the v2 blueprint instead, which should include a config that works.

PR #532, opened 2025-02-12
Feature/min size file upload

aatauil:
## ID
DL-6240

## Description
Currently it is possible to upload files that are 0 bytes. This does not seem very logical. We want to have an option to define a minimum file size and show an error if the file size is lower then that. 

Discussion:
- Do we want to instead make it "minFileSizeKB" and not be consistent with maxFile?
- Do we just create a option that says, "notEmpty" and put the default to 1kb? 

## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Maintanance

## How to setup
No special setup requirements

## How to test
Set the minFileSizeMB to "0.01" (10kb). Try to upload an empty file, you should see the error message. While

### DIFF
--- a/addon/components/au-file-upload.gts
+++ b/addon/components/au-file-upload.gts
@@ -20,6 +20,7 @@ export interface AuFileUploadSignature {
     helpTextDragDrop?: string;
     helpTextFileNotSupported?: string;
     maxFileSizeMB?: number;
+    minFileSizeMB?: number;
     multiple?: boolean;
     onFinishUpload?: (uploadedFile: number, queueInfo: QueueInfo) => void;
     onQueueUpdate?: (queueInfo: QueueInfo) => void;
@@ -87,6 +88,10 @@ export default class AuFileUpload extends Component<AuFileUploadSignature> {
     return this.args.maxFileSizeMB || 20;
   }
 
+  get minFileSizeMB() {
+    return this.args.minFileSizeMB || 0;
+  }
+
   get hasErrors() {
     return this.uploadErrorData.length > 0;
   }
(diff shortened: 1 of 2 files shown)


Windvis:
Should this even be an option? Uploading an empty file does indeed not make sense and do we ever want to configure the minimum size (besides larger than 0)?

aatauil:
@Windvis I guess this option is just more flexible towards the future. I can think of a usecase where a pdf template gets used and that template is by default 100kb, then just checking for 0 is not the solution to check for empty documents.

Update: I have modified the code to use KB instead of MB. setting 0.001mb for 1kb is just bad UX. Theirfor for MIN, kb makes more sense

addon/components/au-file-upload.gts
```diff
@@ -302,6 +312,13 @@ function isValidExtension(
   return validExtensions.includes(extension);
 }
 
-function isValidFileSize(fileSize: number, maximumSize: number): boolean {
-  return fileSize < maximumSize * Math.pow(1024, 2);
+function isValidFileSize(
```
Windvis, line 315: This isn't used anymore, that's also what the linter is complaining about.

Windvis:
> I guess this option is just more flexible towards the future. I can think of a usecase where a pdf template gets used and that template is by default 100kb, then just checking for 0 is not the solution to check for empty documents.

Sure it's more flexible, but chances are high it will never be used. Anyway, I don't have a strong opinion about this, so if you think this is the way to go :+1:.

addon/components/au-file-upload.gts
aatauil: oups that should have been git stashed, somehow got through. Good linter :+1: thumbs

aatauil:
@Windvis Its just an example, another I can think of is a situation where an empty pdf exported from acrobat contains some metadata making the filesize like 15 bytes which would not trigger the "Leeg bestand" error, a dev could then set it to 1kb as a sane minimum

Windvis:
@aatauil I just don't think that setting a minimum file size is the best way to "prevent" those types of issues. It feels very frail, but since it's optional it can't hurt having it either I guess.

Windvis (approved):

Windvis:
Looks good :+1: Could you squash merge it? This can all be 1 commit, imo.

I can do it as well, but then Github adds me as a co-author I think :smile:.

PR #539, opened 2025-05-16
Add new icons from wu and update some existing ones

lagartoverde:
Added some icons that were needed for the rdfa-editor, and updated others to match the ones in webuniversum.
The ones that changed are not fundamentally different from the previous ones so I don't think we have to do anything special


public/icons/html.svg
Windvis, line 1: WU doesn't seem to have a `html` icon, so I'm guessing it has a different name there? I would stick to the WU names so we can deprecate the other (redundant) ones in the future.

Windvis:
Thanks!

PR #541, opened 2025-06-03
add paragraph icon

lagartoverde:
Add paragraph icon that we need for the editor present in webuniversum


Windvis:
It seems the "IconList" component doesn't use the same list of icon names. Could you add it to this list as well? https://github.com/appuniversum/ember-appuniversum/blob/ed956e604b130a6f795dd8edf8d2ca477185091c/tests/dummy/app/components/au-icon-list.js#L5

lagartoverde:
Done :)

Windvis (approved):

PR #578, opened 2026-04-28
add overflow utils

kobemertens:
Adds utility classes for the `overflow` CSS property


Windvis:
Webuniversum only has a [`vl-u-no-overflow` class](https://master--642e92e0cda6c627a0601f07.chromatic.com/?path=/docs/utils-overflow--docs). I wouldn't add all of these just yet.

kobemertens:
Why wouldn't we support all the `overflow` options? Would you then only keep the `overflow--hidden` one? Or do we also keep the `overflow--scroll`? Seems arbitrary to support only 1 or 2 of the possible options?

PR #579, opened 2026-05-11
bulk import webuniversum icons

kobemertens:
Adds all WU icons


Windvis (approved):
Very nice, thanks!

Windvis:
@kobemertens If you have the time (and are interessted in following up on the icon situation). It would be very nice if we can get an overview of icons that are _not_ in Webuniversum, to see if we should deprecate them or not.

Also, I believe Webuniversum also has a list of deprecated icons itself, so we might want to flag those as well on our side.

https://github.com/appuniversum/ember-appuniversum/issues/482

Windvis:
Released as [v4.2.0](https://github.com/appuniversum/ember-appuniversum/blob/master/CHANGELOG.md#v420-2026-05-12) :tada:

PR #580, opened 2026-05-13
Fix the styles exports config

kobemertens:
Files imported using `@use` did not work. This should fix that

### DIFF
--- a/appuniversum/package.json
+++ b/appuniversum/package.json
@@ -17,7 +17,36 @@
   },
   "exports": {
     "./addon-main.js": "./addon-main.cjs",
-    "./styles/*.scss": "./styles/*.scss",
+    "./styles": {
+      "sass": "./styles/_index.scss"
+    },
+    "./styles/*": {
+      "sass": "./styles/_*.scss"
+    },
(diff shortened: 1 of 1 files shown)


appuniversum/package.json
```diff
@@ -17,7 +17,46 @@
   },
   "exports": {
     "./addon-main.js": "./addon-main.cjs",
-    "./styles/*.scss": "./styles/*.scss",
+    "./styles": {
+      "sass": "./styles/_index.scss",
+      "default": "./styles/_index.scss"
```
Windvis, line 22: Any idea why this is needed? I didn't understand the error messages myself.

Windvis (approved):

appuniversum/package.json
kobemertens: Not really, but maybe not all build systems understand the "sass" key? Might even be the case that we also don't need it.

Windvis:
Released as [v4.2.1](https://github.com/appuniversum/ember-appuniversum/blob/master/CHANGELOG.md#v421-2026-05-13)

PR #587, opened 2026-09-07
`AuDataTable` pagination improvements

Windvis:
- styling improvements to match WebUniversum
- option to display pagination links to the first and last pages -> `@showPaginationBoundaryLinks={{true}}`. This might become the default in the future, if most projects want this.
- always display the pagination link, but disable them when needed instead.
- Capitalize the link labels
- fix an issue with the row information numbers
image


Windvis:
It seems we broke something when there is only a single page: https://github.com/lblod/frontend-lpdc/pull/64#pullrequestreview-5197238541

